PR #14, opened 2017-05-02
Hmac authorization

lamalexck:
Ported unsalted/unbounded HMAC authorization test from tpmclient


flihp:
I've taken on the re-write for this test case. Will create a PR when it's ready.

PR #26, opened 2017-05-10
Add back udev rule

lamalexck:


flihp:
The branch associated with this PR was deleted? At least I can't clone it. Either way this is a good catch. I can test the bug and reproduce ('make install' doesn't install this file any more) but I can't pull down your patch.

PR #38, opened 2017-05-22
tpm2-abrmd.8.in: Fix bad default port number for socket TCTI.

flihp:
The default port number used by the simulator is 2321. This is the
default port used by the tpm2-abrmd. The man page was wrong though.
Fixed! See
https://github.com/flihp/tpm2-abrmd/commit/267ec43bddd59c50af3277ae49b9b601423aa4e9#commitcomment-22170495


flihp:
this resolves #38

Issue #45, opened 2017-05-24
make dist fails


flihp:
See comment made on my fork by @liuqun https://github.com/flihp/tpm2-abrmd/commit/90241e49b36d993485a53b737a182aae3ef1dd26

Running `make dist` produces the following output:
```
make  dist-gzip am__post_remove_distdir='@:'
make[1]: Entering directory '/mnt/build/tpm2-abrmd'
make[1]: *** No rule to make target 'dist/tpm2-abrmd.conf.in', needed by 'distdir'.  Stop.
make[1]: Leaving directory '/mnt/build/tpm2-abrmd'
Makefile:3092: recipe for target 'dist' failed
make: *** [dist] Error 2
```

Time for me to read some autotools docs to figure out what 'dist' really means. This is a step on the road to producing real release tarballs.

PR #47, opened 2017-05-25
A temporary solution for issue #45 

liuqun:
It's better to distribute all inneral header files and xml files inside a source tarball. 
Then we can use the following 3 commands to build a gzipped tarball:
```
./bootstrap
./configure
make dist
```
see issue #45


flihp:
Thanks for hashing out a first pass at this @liuqun. I've spent most of today reading through the GNU docs for 'dist' and 'distcheck' and I've gone a slightly different route. I've got a PR together over at #61 if you've got the time to take a look.

Issue #46, opened 2017-05-24
integrate dynamic analysis / valgrind into travis-ci


flihp:
The autotools build already has this stuff enabled. Just need to get travis-ci exercising it.

PR #52, opened 2017-05-28
Check valgrind

flihp:
This series enables valgrind so that `make check-valgrind` runs the unit tests under the valgrind memcheck memory profiler. There are two unit tests, both exercising the Thread lifecycle for derived classes, that had to be removed. These tests both cause a 'critical' error to be thrown up by glib 2.40. They weren't good / valid tests anyways and they should be re-written.


flihp:
this resolves #46

Issue #43, opened 2017-05-23
TctiOption object has 'echo' stuff left over in it


flihp:
The EchoTcti is obsolete. It was some debugging infrastructure that I created early in the development of this daemon to debug the system. It's still hanging around though and has some members polluting the TctiOptions https://github.com/01org/tpm2-abrmd/blob/master/src/tcti-options.h#L51 (and likely beyond). Clean this up.

PR #53, opened 2017-05-29
TctiOptions: Remove final remnants of the EchoTcti.

flihp:
The EchoTcti was some old debugging infrastructure. It was cleaned out
of the daemon a long time back. We still use it in a few test cases so
the code is still around. It's not used in the tpm2-abrmd though.


flihp:
resolves #43

Issue #42, opened 2017-05-23
refactor parse_opts to eliminate side effects


flihp:
[parse_opts ](https://github.com/01org/tpm2-abrmd/blob/master/src/tabrmd.c#L602) has a side effect of creating / instantiating the TctiOptions object. We end up using this object elsewhere so re-ordering the way parse_opts is called could cause weird breakage. The TctiOptions object should be instantiated in the main function.

flihp:
merged in #54

PR #54, opened 2017-05-29
tabrmd: Instantiate the TctiOptions object in main, not parse_opts.

flihp:
The TctiOptions object was instantiated as a side effect of the
parse_opts function. It gets used in the main function immediately after
the call to parse_opts (where the TCTI is instantiated).

It's better form to instantiate this object in main. Then it's passed
into the parse_opts function where the TCTI options are parsed. Then we
use it as a factory to pump out a Tcti object based on the parsed
options.

### DIFF
--- a/src/tabrmd.c
+++ b/src/tabrmd.c
@@ -629,8 +629,6 @@ parse_opts (gint            argc,
         { NULL },
     };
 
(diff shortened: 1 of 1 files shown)


flihp:
this resolves #42

PR #55, opened 2017-05-29
configure: Undefine FORTIFY_SOURCE first, then define it

PeterHuewe:
On my hardened gentoo system _FORTIFY_SOURCE is already defined,
thus it makes sense to undefine it first then define it, not vice
versa.

Signed-off-by: Peter Huewe <peterhuewe@gmx.de>


flihp:
Thanks @PeterHuewe

PR #60, opened 2017-05-29
configure.ac: Use hard coded systemd unit dir not pkgconfig variable.

flihp:
Querying pkg-config for the default here made sense at the time but it
breaks distcheck. This is because distcheck does a DESTDIR install and
the default ignores this.

After thinking about this a bit I'm realizing that using pkg-config also
causes "non-standard" behavior in that it will ignore the expected
behavior when overriding ${libdir}.

This patch simplifies this abit defaulting to ${libdir}/systemd/system.


flihp:
this resolves #60

PR #63, opened 2017-06-01
util: Initialize data, set sane maximum line length in g_debug_buffer.

flihp:
This is intended to prevent the caller from causing a large static
allocation for the 'line' buffer. A line longer than 200 characters will
now produce a warning and print nothing.

We also now allocate the line buffer to 0's since the allocation size is
known at compile time.


flihp:
this change was inspired by some findings from `scan-build`.

Issue #36, opened 2017-05-21
automated static analysis


flihp:
Coverity integrates into travis-ci: https://scan.coverity.com/travis_ci. This code has already gone though static analysis before release (per Intel requirements) but this is something that should happen regularly and the results should be public.

Once #35 is done (travis-ci setup), this should be the next task.

PeterHuewe:
Once it's clang compatible,
Just add clang to the travis compiler list and then 
you can also run scan-build from clang :)
Let me know if you need help

flihp:
Thanks @PeterHuewe. My plan for code analysis goes something like this:
1. dynamic analysis (#52 running unit tests under valgrind)
2. static analysis (coverity)
3. automating integration tests
4. integrating code sanitizers

So clang is pretty far down on that list but it's there. Any cycles you can spare along the way would be much appreciated.

flihp:
Just set up the coverity account for the project. Ran a "first scan". Now waiting for coverity / atlassian folks to verify my association with the project to allow us access to the scan results.

PeterHuewe:
I tried it locally with clang and scan-build and found and fixed some issue (see #57) - only thing that's really blocking clang is #56 (which I worked around locally)

flihp:
Sounds good Peter. I've got the coverity scans working with 3 issues identifed but all are in the test code so they're low priority. I'll get on the clang stuff soon after.

flihp:
See PR #71

PR #71, opened 2017-06-08
Coverity scan

flihp:
Enable automated static analysis on the 'coverity_scan' branch. This includes a few fixups necessary to keep coverity from having a heart attack. There are still a few items to fix in the test code but everything under `src/` got a clean bill of health.


flihp:
This resolves #36

PR #73, opened 2017-06-10
Allow linking against C++

dguerri:


flihp:
That's handy. Thanks @dguerri

Issue #75, opened 2017-06-12
Automate integration test runs


flihp:
Integration tests haven't been included in `make check` or travis-ci runs. The unit tests were easy since they're completely self contained. The integration tests are harder since they require a simulator and tpm2-abrmd instance that is unique to the test. Running these serially isn't too hard, but parallel execution requires the ability to run multiple instances of the simulator / tpm2-abrmd.

Ideally we should
1) start a simulator instance for each integration test
2) start a tpm2-abrmd instance for each integration test, connected 1:1 with the simulator for same test
3) run the test
4) tear down the simulator / tpm2-abrmd pair & cleanup

Running multiple instances of the simulator in parallel requires assigning them unique ports.
Running multiple tpm2-abrmd instances in parallel requires assigning them unique names on the bus. Ideally we would be running these on the session bus to avoid having to run them as root on the system bus.

flihp:
#74 is a step in this process

PR #76, opened 2017-06-13
Integration test log compiler

flihp:
This patch series adds infrastructure that creates a private simulator and tabrmd instance for each integration test before the test is executed. This allows us to add all of the integration tests to the test harness and run them as we do all the unit tests: `make -j$(nproc) check`.


flihp:
this resolves #75

flihp:
Cleaned up the script a bit. Had some debug stuff left in it.

Issue #72, opened 2017-06-09
deal with coverity issues under src/


flihp:
Coverity freaks out if you check a pointer for NULL and call `g_error` to halt the program if said pointer is NULL. Coverity doesn't understand the g_error terminates the program, it just thinks that the program will carry on and then dereference the pointer after the NULL check.

Most of these conditions are after a new object is created. I've been checking for NULL and calling `g_error` if the object isn't instantiated. This seemed like a good idea at the time but it's really redundant. Glib terminates an application if a memory allocation fails so getting a NULL object back isn't something that ever happens. Further it's probably better form to just `g_assert_nonnull` in these scenarios.

Either way the goal here is to placate coverity to get rid of these false positives. The alternative is to teach coverity about this situation using their modeling stuff (https://scan.coverity.com/models) but that will probably take longer / be more work.

PR #80, opened 2017-06-15
coverity: Add model for g_error to deal with false positives.

flihp:
All coverity models should go into this new file. All of our Coverity
false positives come from g_error though so this is all that's in here
for now.


flihp:
This resolves #72

Issue #83, opened 2017-06-19
Please allow multiple contexts to be created by one process on client


dguerri:
Hi Philip, as per our chat:

Given #82 (which is how the code was intended to work), it is not possible to create multiple context per client process, as dbus will always use the same id. 

Please allow creating more that one context per process as it was allowed on the old RM.

flihp:
Yup yup. After our exchange this morning I managed to repro. This was my misunderstanding of how https://developer.gnome.org/gio/stable/GDBusMethodInvocation.html#g-dbus-method-invocation-get-sender works. We'll have to go back to adding an integer component to the connection ID to make the ID unique.

Should have something out pretty quick.

dguerri:
Thanks! ;)

PR #88, opened 2017-06-20
test: Add test case for multiple TCTI connections from a single process.

flihp:
This test currently fails due to a bug in the tabrmd. It's listed in
XFAIL_TESTS until the related bug is squashed.


flihp:
additional work toward a resolution for #83

PR #89, opened 2017-06-20
revert commits that broke client multi-connection & add cmd line seed file option

### LINKED ISSUE #83: Please allow multiple contexts to be created by one process on client

flihp:
This resolves #83


flihp:
Any chance I can get you to test this against your code @dguerri ? This "should" fix it :smile:

dguerri:
It works beautifully, thanks Philip.

flihp:
np @dguerri, thanks for validating

PR #95, opened 2017-06-29
tcti-tabrmd: Gracefully fail if D-Bus connection is not created

martinezjavier:
If creating a D-Bus connection fails, libtcti-tabrmd calls g_error() which
causes an abort() and the process to be terminated. This function should
not be used for expected errors, and failing to create a connection can
happen if for example the tpm2-abrmd.service was not started.

In this case, the program using the library could decide to fallback using
the device TCTI to access the TPM directly or just handle the error better
than being terminated due a SIGABRT signal.

### DIFF
--- a/src/tcti-tabrmd.c
+++ b/src/tcti-tabrmd.c
@@ -472,8 +472,9 @@ tss2_tcti_tabrmd_init_full (TSS2_TCTI_CONTEXT      *context,
         NULL,
         &error);
     if (call_ret == FALSE) {
-        g_error ("Failed to create connection with service: %s",
+        g_warning ("Failed to create connection with service: %s",
                  error->message);
+        return TSS2_TCTI_RC_NO_CONNECTION;
     }
     if (fd_list == NULL) {
         g_error ("call to CreateConnection returned a NULL GUnixFDList");


flihp:
Thanks for this @martinezjavier. This has been on my "TODO" list for way too long. There are a few additional abuses of `g_error` in this code as well (like https://github.com/01org/tpm2-abrmd/blob/master/src/tcti-tabrmd.c#L475) if you feel like taking a look. I should make an issue to track this too ...

martinezjavier:
@flihp you are welcome. Yes, I would guess that most of the g_error() calls don't want an abort but I haven't audited them:

$ git grep g_error | wc -l
140

I thought about changing at least the calls in the tss2_tcti_tabrmd_init_full() function but at the end didn't because I couldn't test the error path (without forcing the error to happen by changing the code) and wasn't sure which TSS2_TCTI_RC_* error code was suitable for each case.

Glib also has a g_printerr() that only prints an error message to stderr without doing an abort.

flihp:
whoops, 140 is a lot :\

The cmocka unit test framework is a great way to inject errors into stuff like the TCTIs. I've done this for most of the "successful" code paths but covering the failure cases would be great. A good example would be unit tests for the GIO D-Bus calls here: https://github.com/01org/tpm2-abrmd/blob/master/test/tss2-tcti-tabrmd_unit.c#L86.

No need to take all 140 on at once. Start with the _init function and go from there ... function by function ... slog slog slog :)

PR #97, opened 2017-06-30
logging: Use G_MESSAGES_DEBUG to enable debug & info logging for sysl…

### LINKED ISSUE #30: syslog handler ignores G_MESSAGES_DEBUG
When the tpm2-abrmd is started with `--logger=syslog` the log level is dialed up all the way regardless of the state of G_MESSAGES_DEBUG. This is a bug in the `syslog_log_handler` function.

flihp:
…og handler.

This isn't "the right way" to do this but it's better than the current
state of affairs. This patch disables info and debug messages for the
syslog handler by default. If the G_MESSAGES_DEBUG environment variable
is set to "all" then info and log messages are enabled.

This resolves #30 


dguerri:
<3 thanks

PR #103, opened 2017-07-24
switched signal handling from sigaction() to g_unix_signal_add()

mgerstner:
Calling g_main_loop_is_running() from an asynchronous signal handler
results in undefined behaviour. See documentation of
g_unix_signal_source_new():

https://developer.gnome.org/glib/stable/glib-UNIX-specific-utilities-and-integration.html#g-unix-signal-source-new

This commit switches over to glib based signal handling. This also
eliminates the need for the global variable holding the pointer to the
main loop data structure.

### DIFF
--- a/src/tabrmd.c
+++ b/src/tabrmd.c
@@ -28,6 +28,7 @@
 #include <fcntl.h>
 #include <gio/gio.h>
 #include <glib.h>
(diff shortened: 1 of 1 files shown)


flihp:
This looks great @mgerstner. Thanks!

PR #124, opened 2017-08-03
test: Abort int-log-compiler.sh with error codes if required bins not…

flihp:
… found.

The simulator, tabrmd and test executables are required for this script
to do anything meaningful. Better to fail early with a clear message
indicating what went wrong.


flihp:
this resolves #124

PR #125, opened 2017-08-04
Security and Robustness Fixes

mgerstner:
Found during internal security audit for inclusion for tpm2-abrmd in SUSE Linux distributions.


flihp:
Looks good @mgerstner. Thanks for taking the time to fix this stuff up.

PR #140, opened 2017-08-15
fix sapi lib inclusion

williamcroberts:
When building tabrmd with a custom install location for sapi,
it couldn't find tpm20.h header file, as it was missing the
sapi prefix.

Correct this by adding it.

Example configure:
./configure --prefix=`realpath ~/tmp` PKG_CONFIG_PATH=`realpath ~/tmp/lib/pkgconfig`

### DIFF
--- a/src/tpm2-command.h
+++ b/src/tpm2-command.h
@@ -28,7 +28,7 @@
 #define TPM2_COMMAND_H
 
 #include <glib-object.h>
-#include <tpm20.h>
+#include <sapi/tpm20.h>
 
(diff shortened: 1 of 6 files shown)


williamcroberts:
I'm actually seeing this on the tools build, which installs to the default locations: https://travis-ci.org/01org/tpm2-tools/jobs/264786066

flihp:
LGTM. I've created an issue in the tools project to discuss the source of the breakage you're seeing there.

PR #142, opened 2017-08-16
CHANGELOG.md: Update for diff between 1.0.0 and 1.1.0 releases.

flihp:


flihp:
I should note that this is fixing up something that was never done for the 1.1.0 release (my mistake). The patches / commits between 1.0.0 and 1.1.0 can be dumped with: `git log 1.0.0..1.1.0` for anyone who wants to check the accuracy of the log.

PR #151, opened 2017-08-21
build against tss master branch

williamcroberts:
src/access-broker.c:446:13: error: ‘buffer’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
             free (buffer);

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ install:
   - cmake ../ -DCMAKE_INSTALL_PREFIX=${PREFIX} -DCMAKE_BUILD_TYPE=Release
   - make -j$(nproc) install
   - popd
-  - git clone -b 1.x --single-branch https://github.com/01org/tpm2-tss.git
+  - git clone -b master --single-branch https://github.com/01org/tpm2-tss.git
   - pushd tpm2-tss
   - ./bootstrap
   - ./configure --prefix=${PREFIX}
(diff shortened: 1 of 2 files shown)


flihp:
Not sure why the checks failed here but your patches look fine. You OK with me squashing these all into a single commit to keep the source tree building in the intermediate states? This is to keep bisects working on the off chance someone needs to bisect.

flihp:
The build fail is from some library BS. Something about the dependency from libsapi to libmarshal is broken when the libraries aren't installed with DESTDIR=/. Will do some digging on this.

williamcroberts:
@flihp Yeah I tried to figure out what that library error was about, but I never got anywhere. You can squash this into a single commit if you'd like.

williamcroberts:
On #163 without travis update (so likely will fail again)

flihp:
thx @williamcroberts, I've been working on a topic branch with this stuff squashed down. Still grinding down the library issue. It's driving me nuts.

flihp:
I've got it building with travis & clang but travis & gcc is convinced that there are missing symbols. I can build it locally with both but travis + gcc is still broken.

williamcroberts:
Im building with these patches on tools travis... but im not doing all the checking

flihp:
yeah it's just one of the test cases won't build under GCC: https://github.com/01org/tpm2-abrmd/pull/167/commits/79672abcbdb39063635469efb22f798160b97b09#r136211207

PR #153, opened 2017-08-22
Refuse running the daemon as root by default

mgerstner:
A little suggestion, refusing to run as root by default. Maybe doing this via the default D-Bus service configuration is cleaner. On the other hand this approach is more explicit and noticeable.

Also one minor bugfix is included I stumbled upon.

### DIFF
--- a/src/tabrmd.c
+++ b/src/tabrmd.c
@@ -746,6 +746,9 @@ parse_opts (gint            argc,
           options->prng_seed_file },
         { "version", 'v', G_OPTION_FLAG_NO_ARG, G_OPTION_ARG_CALLBACK,
           show_version, "Show version string" },
+        { "allow-root", 0, 0, G_OPTION_ARG_NONE,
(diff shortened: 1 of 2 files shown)


flihp:
Not sure how I feel about the new command line option & default behavior. The bug fix is a good one though so I'm going to pull that into the bug fix release I'm currently putting together. I'll create an issue to track the new behavior for consideration in the next feature release.

flihp:
Thanks for leaving the branch up while I finished off some other "stuff". I've been working mostly on the 1.x branch preparing for another minor release. Since this last half of this PR changes default behavior I'll pull it into the master branch in preparation for a major version release that will hopefully be timed with a major version release of the TSS.

flihp:
I've rebased this on the latest master, added a short option / flag (which may be changed before the next major release) and created a new PR: https://github.com/01org/tpm2-abrmd/pull/191

PR #160, opened 2017-08-23
build: Add options to the configure script for the systemd preset file.

flihp:
This allows the caller to specify the default install path for the
preset file, as well as the enable / disable state for the service unit.
By default the install path is `${libdir}/systemd/service-preset`. This
can be overriden using the --with-systemdpresetdir=DIR option. By
default the tpm2-abrmd preset will enable the daemon. It may be disabled
by default using the `--with-systemdpresetdisable` option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,5 @@ test/tcti-tabrmd_read-write
 test/integration/*
 !test/integration/*.c
 !test/integration/*.h
+dist/*.preset
 dist/*.service
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -103,18 +103,33 @@ following configure option:
 --with-dbuspolicydir=/etc/dbus-1/system.d
 ```
 
-### Systemd Uint: `--with-systedsystemunitdir`
+### Systemd
 In most configurations the `tpm2-abrmd` daemon should be started as part of
-the boot process. To enable this we provide a systemd unit. By default the
-build installs this file to `${libdir}/systemd/system`. Just like D-Bus
-the location of unit files is distro specific and so you may need to
-configure the build to install this file in the appropriate location.
+the boot process. To enable this we provide a systemd unit as well as a
+systemd preset file.
+
+#### Systemd Uint: `--with-systedsystemunitdir`
+By default the build installs this file to `${libdir}/systemd/system. Just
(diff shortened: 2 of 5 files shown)


yunyings:
I cloned your repo of flihp/tpm2-abrmd, and checkout to branch systemd-preset. Made a source tarball from the code base(named it as tpm2-abrmd.1.1.0.tar.gz), and updated SPEC file to add scriptlets below:
```
# reload udev rules and apply the new rule
udevadm control --reload-rules
udevadm trigger
# make dbus-daemon re-read its configuration file
pkill -HUP dbus-daemon
# make systemd reload its configuration
systemctl daemon-reload
# start tpm2-abrmd service
systemctl enable tpm2-abrmd
```

Then I re-created tpm2-abrmd RPM and tried it on RHEL 7.3. Results:
1. tpm2-abrmd daemon does not get started after package installation.
2. tpm2-abrmd daemon does not start on system reboot after package installation.
3. manually start daemon but "sudo tpm2-abrmd" works fine. tpm2_listpcrs shows PCR registers as expected.

Here I attached the RPMs that I used for testing and tpm2-abrmd SPEC file. Philp please feel free to try on RHEL 7.x.

Two questions for you Philip: 
1. Is tpm2-abrmd daemon/service auto start works on other OSVs(like Ubuntu) after package installation and system reboot? Do you have such info?

2. Suppose you must have verified the daemon auto start feature for upstream. Which distribution are you using for verification, Ubuntu? Does it work by normal configure & make & install & post-install steps as shows in INSTALL.md?

[tpm2-pakcages.zip](https://github.com/01org/tpm2-abrmd/files/1251433/tpm2-pakcages.zip)

flihp:
I don't have access to a RHEL license @yunyings so don't think I'll be able to repro for you. The systemd stuff works on my dev system (Debian) and the OpenEmbedded imagines that I use for testing on a Minnowboard. Neither of these two distros use these preset files though.

I'll try to dig up another system to test using Fedora, probably the closest I'll be able to come to a RHEL box (unless you think Centos is better?). Till then I'm just going to merge this and make the next release since it's just build stuff. We can sort this out for the next release.

yunyings:
Fedora should be good to check this issue. I will look into the preset stuff too to see what I can do.

PR #162, opened 2017-08-25
build: Check for compatible SAPI version @ configure time.

flihp:
The upstream TSS master branch now has API / ABI incompatible changes
with the 1.x stable releases. The tabrmd releases are maintaining
compatibility with the 1.x TSS until a 2.x release is made.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,7 @@ PKG_CHECK_MODULES([DBUS], [dbus-1])
 PKG_CHECK_MODULES([GIO], [gio-unix-2.0])
 PKG_CHECK_MODULES([GLIB], [glib-2.0])
 PKG_CHECK_MODULES([GOBJECT], [gobject-2.0])
-PKG_CHECK_MODULES([SAPI],[sapi])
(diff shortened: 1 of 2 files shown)


flihp:
We should probably do another bug fix release with this change to keep people from shooting themselves in the foot.

configure.ac
```diff
@@ -29,7 +29,7 @@ PKG_CHECK_MODULES([DBUS], [dbus-1])
 PKG_CHECK_MODULES([GIO], [gio-unix-2.0])
 PKG_CHECK_MODULES([GLIB], [glib-2.0])
 PKG_CHECK_MODULES([GOBJECT], [gobject-2.0])
-PKG_CHECK_MODULES([SAPI],[sapi])
+PKG_CHECK_MODULES([SAPI],[sapi < 2.0.0])
```
liuqun: Consider to leave some warning messages to 3rd-party developers when sapi >= 2.0.0 is found
(instead of such a fatal error with no detailed messages, which might disturb anyone trying to build tabrmd with an unstable version of libsapi)

PR #166, opened 2017-08-30
Update INSTALL.md

rtkypoeluev:
Typo in --with-systedsystemunitdir control, should be "--with-systemdsystemunitdir"


flihp:
Good catch @rtkypoeluev. Thanks!

PR #167, opened 2017-08-30
Master tracks master destdir

flihp:
This was a bit of a slog. The additional libraries from the 2.0.0 TSS combined with the DESTDIR install required we figure out how to handle the discrepancy between the DESTDIR path and the `dependency_libs` entry in the `*.la` files. Additionally this PR removes a test case from the build. This test case causes gcc linking issues while clang has no problem. It's probably best to make the disabling of this test configurable so devs can run it locally since it works fine on my dev system, only travis seems to be effected.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,7 +30,6 @@ TESTS_UNIT = \
     test/session-entry_unit \
     test/test-skeleton_unit \
     test/tcti-echo_unit \
-    test/tcti-options_unit \
     test/thread_unit \
     test/tpm2-command_unit \
     test/tpm2-response_unit \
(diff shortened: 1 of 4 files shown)


Makefile.am
```diff
@@ -352,11 +351,6 @@ test_tcti_socket_unit_LDFLAGS  = -Wl,--wrap=InitSocketTcti
 test_tcti_socket_unit_SOURCES  = test/tcti-socket_unit.c
 endif
 
-test_tcti_options_unit_CFLAGS  = $(UNIT_AM_CFLAGS)
-test_tcti_options_unit_LDADD   = $(CMOCKA_LIBS) $(GLIB_LIBS) $(GOBJECT_LIBS) \
-    $(libutil) $(libtcti_echo) $(TCTI_DEVICE_LIBS) $(TCTI_SOCKET_LIBS)
-test_tcti_options_unit_SOURCES = test/tcti-options_unit.c
```
flihp: This is the test case that I'm struggling with @williamcroberts. I can't explain why but with the 2.0 TSS this test case fails to build for GCC in travis. Clang builds and works fine, as does GCC 4.8.4 (same version as in travis) on my dev system.

I may have been looking at this test code for too long so the problem may be obvious to a fresh set of eyes. For the short term I'm going to disable the test.

liuqun:
I think there might be some existing files on your local dev system, which does exist on travis.
Or the version of libtool may be a outdated buggy version on travis.
Check libtool/gcc default search path for *.so *.a on travis might help.
Good Luck!

PR #181, opened 2017-09-18
Merge 1.x to master

flihp:
This merge got a bit borked with duplicate commits between 1.x and master. Not sure how I made this happen but it should just be some extra noise that can be ignored.


flihp:
Github doesn't like PRs for weird merges like this. It wants to create its own merge commit on top of the existing merge commit in the PR. I'm going to let the travis check run and manually push the merge and close this PR once it passes.

PR #183, opened 2017-09-22
Tcti fd list unref

flihp:


flihp:
Breakage is caused by upstream TSS 1.x build. Have a patch up there to fix: https://github.com/01org/tpm2-tss/pull/550

flihp:
upstream fix merged, this PR should be good to go

PR #185, opened 2017-09-25
1.x fd removal refactor

flihp:


src/command-source.c
```diff
@@ -156,39 +173,161 @@ command_source_get_property (GObject      *object,
...
+    g_hash_table_foreach (self->socket_to_source_data_map,
+                          command_source_source_cancel,
+                          NULL);
+    g_debug ("%s: clearing socket_to_source_data_map with g_hash_table_unref",
+             __func__);
+    g_clear_pointer (&self->socket_to_source_data_map, g_hash_table_unref);
+    g_debug ("%s: done", __func__);
+}
 
 /* Not doing any sanity checks here. Be sure to shut things downon your own
```
lugong: miss a space between down and on.

src/connection.c
```diff
@@ -214,7 +208,11 @@ connection_new (gint       *client_fd,
         *client_fd = 0;
         return NULL;
     }
```
lugong: GSocket is used to monitor G_IO_IN event, and this replaces the function select().
How about making use of the following functions to construct a GIOStream structure: g_unix_input_stream_new(), g_unix_output_stream_new() and g_simple_io_stream_new()? This GIOStream structure is used to read/write data from/to client and consistent with GTlsServerConnection/GTlsClientConnection.

src/command-source.c
flihp: good catch!

src/connection.c
flihp: Do you know of a mechanism for getting G_IO_IN events from GInputStreams? I looked and was only able to find this mechanism in GSockets. Maybe I just didn't look hard enough.

Either way my goal is to do exactly what you suggest. I think that's the next step after this gets integrated and tested a bit.

src/command-source.c
flihp: oh wow that's an old comment, not from this patch. It's not accurate any more. I'll fix this up in a separate commit.

src/connection.c
flihp: Ah I hadn't found the GPollableInputStream yet: https://developer.gnome.org/gio/stable/GPollableInputStream.html. Pretty sure this is what we need.

lugong: Yes, agree.
With GPollableInputStream and GPollableOutputStream, it is feasible to handle read/write events in a stream level rather than the socket layer. The G_IO_IN events is no longer needed to generate an event source.

PR #190, opened 2017-09-28
Command: Replace use of GSocket with GIOStream.

flihp:
This change touches a number of other objects. The CommandSource and
ResponseSink objects now use GInputStream and GOutputStreams
respectively. The ConnectionManager object is also updated to allow
lookups from the CommandSource using the GInputStream associated with
the Connection.

The util module is also updated to use GInputStream and GOutputStream
objects in place of the GSockets.

Finally all of the unit tests that handle Connection objects or do I/O
using the util functions had to be updated.


flihp:
This work was waylaid by my initial effort using a GObject that wasn't introduced to GLib till version 2.44. The travis-ci images are based on an Ubuntu version that ships version 2.42. Wasn't a huge amount of work to find another way to accomplish the task but it did slow this patch down by a few days.

PR #191, opened 2017-09-29
tabrmd: refuse to run as root by default

flihp:
To avoid unwise configurations or configuration errors, refuse to run as
root by default. A new boolean command line switch allows to override
this behaviour by explicitly adding that switch to the command line.


flihp:
This repackages a previous PR from @mgerstner with a rebase and minor edit to add a short option / flag

mgerstner:
Thanks for considering this change.

Issue #189, opened 2017-09-28
DBus name acquisition / loss reasons


flihp:
Currently if the daemon isn't able to acquire the requested name on DBus the daemon shuts down but doesn't give the caller an indication as to why. The skilled debugger will turn to DBus and get debugging info there (where the rejection messages are very clear), but mostly I'm getting emails asking "why doesn't this work".

To fix this we should get some better output when the daemon can't acquire the requested DBus name. The GIO docs however make getting at this information sound difficult (requires tracking which callbacks have been called: https://developer.gnome.org/gio/stable/gio-Owning-Bus-Names.html#g-bus-own-name) so there may be more work here than would be expected.

flihp:
PR for this is up: #193

PR #193, opened 2017-10-01
IpcFrontendDbus: Give better warning when name can't be acquired.

flihp:
The event triggered when the name is lost doesn't provide a reason
as to why. As a result the warning displayed when this happened left
a lot of users guessing as to why. According to the GIO docs, if we
try to acquire a name on DBus and get the name lost event without
first getting the name acquired event then the we have a problem with
permissions in the DBus config.

https://developer.gnome.org/gio/stable/gio-Owning-Bus-Names.html#g-bus-own-name

### DIFF
--- a/src/ipc-frontend-dbus.c
+++ b/src/ipc-frontend-dbus.c
@@ -115,7 +115,9 @@ ipc_frontend_dbus_get_property (GObject    *object,
 }
 static void
 ipc_frontend_dbus_init (IpcFrontendDbus *self)
-{ /* n00p */ }
+{
+    self->dbus_name_acquired = FALSE;
+}
 /*
  * Dispose method where where we free up references to other objects.
  */
@@ -607,6 +609,7 @@ on_name_acquired (GDBusConnection *connection,
         &error);
     if (ret == FALSE)
         g_warning ("failed to export interface: %s", error->message);
+    self->dbus_name_acquired = TRUE;
 }
 /*
  * This is a signal handler of type GBusNameLostCallback. It is
@@ -619,8 +622,17 @@ on_name_lost (GDBusConnection *connection,
               const gchar     *name,
               gpointer         user_data)
 {
-    g_info ("on_name_lost: %s", name);
+    g_debug ("%s: %s", __func__, name);
     IpcFrontend *ipc_frontend = IPC_FRONTEND (user_data);
+    IpcFrontendDbus *self = IPC_FRONTEND_DBUS (user_data);
+
+    if (self->dbus_name_acquired == FALSE) {
+        g_critical ("Failed to acquire DBus name %s. UID %d must be "
+                    "allowed to \"own\" this name. Check DBus config.",
+                    name, getuid ());
+    } else {
+        self->dbus_name_acquired = FALSE;
+    }
 
     ipc_frontend_disconnected_invoke (ipc_frontend);
 }
--- a/src/ipc-frontend-dbus.h
+++ b/src/ipc-frontend-dbus.h
@@ -51,6 +51,7 @@ typedef struct _IpcFrontendDbus
     gchar             *bus_name;
     GBusType           bus_type;
     /* private data */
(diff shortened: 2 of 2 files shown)


flihp:
this resolves #189

src/ipc-frontend-dbus.c
```diff
@@ -607,6 +609,7 @@ on_name_acquired (GDBusConnection *connection,
         &error);
     if (ret == FALSE)
         g_warning ("failed to export interface: %s", error->message);
+    self->connection_state = CONNECTION_STATE_CONNECTED;
```
martinezjavier: I think this state is misleading. Is not about establishing or not a connection, but about obtaining or not a well known name.

If a connection can't be made to the bus, then the on_name_lost callback handler will be called with a NULL GDBusConnection (and the bus_acquired_handler callback will never be called).

So if this is about making sure that a connection was made, then this has to be set in on_bus_acquired. But the intention is to know if the on_name_acquired callback was really called (since if the name can't be acquired, then on_name_acquired will never be called, just bus_acquired_handler followed by name_lost_handler).

That why I would instead use something like BUS_NAME_ACQUIRED. In fact, I don't think you need an enum, just a bool name_acquired should be enough.

src/ipc-frontend-dbus.c
```diff
@@ -619,9 +622,17 @@ on_name_lost (GDBusConnection *connection,
               const gchar     *name,
               gpointer         user_data)
 {
-    g_info ("on_name_lost: %s", name);
+    g_debug ("%s: %s", __func__, name);
     IpcFrontend *ipc_frontend = IPC_FRONTEND (user_data);
+    IpcFrontendDbus *self = IPC_FRONTEND_DBUS (user_data);
+
+    if (self->connection_state == CONNECTION_STATE_DISCONNECTED) {
+        g_warning ("Failed to acquire DBus name %s. UID %d must be "
```
martinezjavier: Wonder if this shouldn't be g_critical instead since is more than a warning (the daemon will not start after this)?

src/ipc-frontend-dbus.c
```diff
@@ -619,9 +622,17 @@ on_name_lost (GDBusConnection *connection,
...
+                   "allowed to \"own\" this name. Check DBus config,",
```
martinezjavier: I don't think the comma at the end of the message log is correct. Otherwise users will wonder if there was some part of the sentence that wasn't printed.

flihp: this was a typo, good catch

src/ipc-frontend-dbus.c
flihp: Reasonable. The daemon will shutdown after this so the code treats the condition as 'critical' since it means there's no reason for the daemon to try to do anything but shutdown if this happens.

src/ipc-frontend-dbus.c
flihp: This is a good simplification. Will give it a try to see how it looks in practice.

src/ipc-frontend-dbus.c
```diff
@@ -619,8 +622,15 @@ on_name_lost (GDBusConnection *connection,
...
-    g_info ("on_name_lost: %s", name);
+    g_debug ("%s: %s", __func__, name);
     IpcFrontend *ipc_frontend = IPC_FRONTEND (user_data);
+    IpcFrontendDbus *self = IPC_FRONTEND_DBUS (user_data);
+
+    if (self->dbus_name_acquired == FALSE) {
+        g_critical ("Failed to acquire DBus name %s. UID %d must be "
+                    "allowed to \"own\" this name. Check DBus config.",
+                    name, getuid ());
+    }
```
flihp: ugh, looks like I forgot to reset the flag to FALSE here if it's not already

flihp:
Of course. Thanks for taking the time to review this for us. More eyes == better code!

PR #195, opened 2017-10-02
Add integration test for session save / load across connections.

flihp:
This is a feature needed by the tools (and likely many other). Currently
the test fails as expected since the daemon doesn't suppor the feature
... yet.

### DIFF
--- a/test/integration/session-save-across-connections.int.c
+++ b/test/integration/session-save-across-connections.int.c
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2017, Intel Corporation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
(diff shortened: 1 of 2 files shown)


flihp:
Associated branch to add support for this in the daemon is here: https://github.com/flihp/tpm2-abrmd/tree/1.x-session-save-across-connection-test.

The code on this branch already gets this test case to pass. What it doesn't do yet is any sort of cleanup so it could leave sessions around in the TPM when the daemon terminates etc. I'll get on that cleanup soon. Till then some verification that the code does what we want it to do would be much appreciated.

flihp:
related: https://github.com/01org/tpm2-tools/issues/532

test/integration/session-save-across-connections.int.c
```diff
@@ -0,0 +1,116 @@
...
+ */
+#include <inttypes.h>
+#include <glib.h>
+#include <stdio.h>
+
+#include <sapi/tpm20.h>
+
+#include "common.h"
+#include "test-options.h"
+#include "context-util.h"
```
williamcroberts: typically when people break these up in blocks like this, they do:
std headers

system headers

internal headers

And sort them by name.

So glib would be system header... and the blocks would need to be sorted.

test/integration/session-save-across-connections.int.c
```diff
@@ -0,0 +1,116 @@
...
+    }
+    g_info ("Got SAPI context: 0x%" PRIxPTR, (uintptr_t)sapi_context);
+    /* create an auth session */
+    g_info ("Starting unbound, unsaulted auth session");
+    rc = start_auth_session (sapi_context, &session_handle);
+    if (rc != TSS2_RC_SUCCESS) {
+        g_error ("Tss2_Sys_StartAuthSession failed: 0x%" PRIxHANDLE, rc);
+    }
+    g_info ("StartAuthSession for TPM_SE_POLICY success! Session handle: "
+            "0x%08" PRIx32, session_handle);
```
williamcroberts: Wouldn't this be PRIxHANDLE?

williamcroberts:
@flihp actually the continuation session code is found on this branch: https://github.com/01org/tpm2-abrmd/compare/master...flihp:1.x-session-save-across-connections

Not:
https://github.com/flihp/tpm2-abrmd/tree/1.x-session-save-across-connection-test

At least when I checked, https://github.com/flihp/tpm2-abrmd/tree/1.x-session-save-across-connection-test was missing this commit: https://github.com/01org/tpm2-abrmd/commit/7b92d3b92b46581aafbb9da8e943624e54cb8d8f

test/integration/session-save-across-connections.int.c
flihp: ACK. Not something I've been particularly strict about but it's never too late to fix this stuff up.

test/integration/session-save-across-connections.int.c
flihp: I've been inconsistent with the use of this macro. It was a wellintentioned effort to make the code more readable and keep me from having to remember to add the `08` padding / width stuff. I'll fix this one up and add a ticket to track moving all of these test cases over.

flihp:
Whoops. Good catch. That was a copy / paste mistake.

test/integration/session-save-across-connections.int.c
flihp: Would be interested to know what you thought about something like PRIxHANDLE in general? I added it after forgetting the padding / width stuff when dumping handles and confusing myself while reading debug output. I copied the style from `inttypes` which seemed like a good idea at the time but now looks like it may cause confusion as to where it's defined.

williamcroberts: I don't have a very strong opinion of it either way

flihp:
resolves #198

Issue #188, opened 2017-09-27
tpmdd vs. tpm2-abrmd


danglus:
Hi,
  What are the relationships between the kernel resource manager (https://sourceforge.net/p/tpmdd/mailman/tpmdd-devel/) and this project?
  What are the guidelines for using one vs. the other?

(this is a followup to https://github.com/01org/tpm2-tss/issues/552#issuecomment-332249601)

   Thanks,
Dan

flihp:
Good question. Jarkko and I did a talk about this very subject @ LPC 2 weeks back but the talk wasn't recorded and the LPC website doesn't have a link to the slide deck (https://www.linuxplumbersconf.org/2017/ocw/proposals/4675) so I'll summarize here a bit.

The current implementations are mostly equivalent. They both provide isolation between objects & sessions created by different connections. Given this, the reason we have both still is that we have customers running with kernels going back to the 3.x series and the kernel RM is only available on very recent kernels. So the user space RM will be around at least till everyone is using the kernel RM.

For how we get to the ideal future of a single RM in the kernel: my plan is to prototype various features in user space as a way to get them tested. There's a lot of stuff in the related TCG spec that we haven't yet implemented and we all agree that trying to put features into the kernel before we really believe them to be useful, understand how they work, and how they're going to be used by applications is a bad idea.

A good example of this are the asynchronous portions of the SAPI. Right now with the kernel RM you can use the async API but it won't really be async (calls to functions that should be async will block) since the kernel doesn't supply user space with an async / polling I/O interface. For the short term, if you want to use the SAPI in an event driven I/O framework you will only get async I/O from the user space resource manager. In the long run though, *if* this feature is important to folks we can work with the kernel folks to add support there. My plan is to treat future features in the same way.

Hope this helps clear things up a bit.

martinezjavier:
@flihp that's very good information, do you think that could add something like that in the docs or the project wiki? (instead having it buried in an issue likely to be closed soon).

Also it would be great if you can share your LPC talk slides, maybe in the mailing list.

flihp:
I'm trying to get the LPC organizers to expose the link through their site. If that fails I'll host them myself.

Getting data on the wiki is always a good idea. I'm planning to dump some hours into documentation once we work out the session mgmt stuff for the tools policy mechanism.

martinezjavier:
@flihp great, thanks!

martinezjavier:
[...]

PR #201, opened 2017-10-03
README.md: Add info about tabrmd vs in-kernel RM.

flihp:
This adds details about the migration from the user space tabrmd to use
of the in-kernel implementation. This was the subject of the talk we
gave @ LCP this year so we include a link to the slides as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,45 @@ tabrmd on the system bus as this is the most common configuration.
 
 Check out the man page TCTI-TABRMD(7) and TSS2_TCTI_TABRMD_INIT(3).
 
+## tpm2-abrmd vs in-kernel RM
+The current implementations are mostly equivalent with a few differences.
+Both provide isolation between objects & sessions created by different
+connections which is the core functionality required by applications. The
+reason we have both is that the in-kernel RM was only added very recently
+(4.12) and we have TPM2 users in environments with kernels going back to the
+3.x series. So the user space RM will be around at least till everyone is
+using the kernel RM.
+
+For the short term we're recommending that developers stick to using the
+tabrmd as the default to get the most stable / widest possible support.
+If you structure your code properly you'll be able to switch in / out TCTI
+modules with relative ease and migrating to the in-kernel RM should be pretty
+painless. Eventually, all of the required features will end up in the kernel
+RM and it will become the default.
+
+How we get to the ideal future of a single RM in the kernel: our current plan
+is to prototype various features in user space as a way to get them tested /
+validated. There's a lot of stuff in the related TCG spec that we haven't yet
+implemented and we all agree that it's generally a bad ideal to to put
+features into the kernel before we:
+1. understand how they work
+2. how they're going to be used by applications
+3. agree we want the feature at all
+
+A good example of this are the asynchronous portions of the SAPI. Right now
+with the kernel RM you can use the async API but it won't really be
+asynchronous: Calls to functions that should be async will block since the
+kernel doesn't supply user space with an async / polling I/O interface. For
+the short term, if you want to use the SAPI in an event driven I/O framework
+you will only get async I/O from the user space resource manager. In the long
+run though, if this feature is important to our users, we can work to upstream
+support to the in-kernel RM. The plan is to treat future features in the same
+way.
+
+This was the subject of a talk that was given @ the Linux Plumbers Conference
+2017:
+http://linuxplumbersconf.com/2017/ocw//system/presentations/4818/original/TPM2-kernel-evnet-app_tricca-sakkinen.pdf
+
 # Related Specifications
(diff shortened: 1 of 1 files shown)


flihp:
This resolves #188

README.md
```diff
@@ -33,6 +33,33 @@ tabrmd on the system bus as this is the most common configuration.
```
martinezjavier: There's a typo in the commit message s/LCP/LPC. And also I would say change LPC this year to LPC 2017. It saves you one level of indirection on having to look at the commit date to know what year that is.

README.md
```diff
@@ -33,6 +33,33 @@ tabrmd on the system bus as this is the most common configuration.
 
 Check out the man page TCTI-TABRMD(7) and TSS2_TCTI_TABRMD_INIT(3).
 
+## tpm2-abrmd vs in-kernel RM
+The current implementations are mostly equivalent with a few differences.
+Both provide isolation between objects & sessions created by different connections which is the core functionality required by applications.
+The reason we have both is that the in-kernel RM was only added very recently and we have TPM2 users in environments with kernels going back to the 3.x series.
+So the user space RM will be around at least till everyone is using the kernel RM.
```
martinezjavier: I would add that the in-kernel RM was included in 4.12. That way users reading this document will know exactly what are the requirements to use the kernel RM.

README.md
```diff
@@ -33,6 +33,33 @@ tabrmd on the system bus as this is the most common configuration.
...
+If you structure your code properly you'll be able to switch in / out TCTI modules with relative ease and migrating to the in-kernel RM should be pretty painless.
+Eventually, all of the required features will end up in the kernel RM and it will become the default.
+
+How we get to the ideal future of a single RM in the kernel: our current plan is to prototype various features in user space as a way to get them tested / validated.
+There's a lot of stuff in the related TCG spec that we haven't yet implemented and we all agree that trying to put features into the kernel before we:
+1. understand how they work
+2. how they're going to be used by applications
+3. agree we want the feature at all
+
+is generally a bad idea.
```
martinezjavier: I would move this last sentence above the three points mentioned, i.e: "we all agree that's generally a bad idea trying to put features into the kernel before we"

Otherwise if people are not reading carefully, they may have to go back and re-read the sentence before the points to remember what was a bad idea.

README.md
```diff
@@ -33,6 +33,33 @@ tabrmd on the system bus as this is the most common configuration.
...
+
+A good example of this are the asynchronous portions of the SAPI.
+Right now with the kernel RM you can use the async API but it won't really be asynchronous:
+Calls to functions that should be async will block since the kernel doesn't supply user space with an async / polling I/O interface.
+For the short term, if you want to use the SAPI in an event driven I/O framework you will only get async I/O from the user space resource manager.
+In the long run though, if this feature is important to our users, we can work to upstream support to the in-kernel RM.
+The plan is to treat future features in the same way.
```
martinezjavier: I would also mention as an example the session context flush that we are discussing now. Specially since IIUC the fact that the in-kernel RM flushes the sessions when the connection (file descriptor) is closed is a design choice, so this will be a difference in behaviour between the user-space and kernel-space resource managers.

README.md
williamcroberts: +1

README.md
williamcroberts: +1

README.md
flihp: I think I'm physically incapable of typing LPC given the number of times I've typed out LCP (launch control policy)

README.md
flihp: This makes sense but I'm going to hold off adding it till the feature is merged. I used the async stuff as an example since it already works in past releases and I'm working on a write up describing how to use it with  GIO event sources.

flihp: Probably best to keep a structured list of places where the two implementations differ.

README.md
```diff
@@ -33,6 +33,45 @@ tabrmd on the system bus as this is the most common configuration.
...
+support to the in-kernel RM. The plan is to treat future features in the same
+way.
+
+This was the subject of a talk that was given @ the Linux Plumbers Conference
+2017:
+http://linuxplumbersconf.com/2017/ocw//system/presentations/4818/original/TPM2-kernel-evnet-app_tricca-sakkinen.pdf
+
 # Related Specifications
 * [TPM2 Software Stack Access Broker and Resource Manager](http://www.trustedcomputinggroup.org/wp-content/uploads/TSS-TAB-and-Resource-Manager-00-91-PublicReview.pdf)
 * [TPM2 Software Stack System API and TPM2 Command Transmission Interface](http://www.trustedcomputinggroup.org/wp-content/uploads/TSS-system-API-01.pdf)
```
tstruk: That's an invalid link. I think we want this one:
https://trustedcomputinggroup.org/wp-content/uploads/TSS_TCTI_v1.0_r04_Public-Review.pdf

README.md
martinezjavier: @flihp right, that makes sense to me.

README.md
flihp: Thanks @tstruk. I'll create a new PR to fix that link since it's not part of what we're reviewing here.

Issue #203, opened 2017-10-04
daemon dies on system bus


flihp:
I've just run into a weird issue where the daemon appears to be getting kicked off the dbus system bus when clients execute a command, but it works fine on the session bus. I'm collecting data still and trying to isolate the change that introduced the issue (if there is one).

flihp:
The failure occurs after tabrmd switched from using pipes to sockets. Still the issue only happens when the daemon is on the system bus. Session bus works fine. This makes me think it's not in the code here but something to do with the difference between the two bus types. Not sure how sockets vs pipes plays in to this.

flihp:
Oh SELinux, how I've missed you. Problem goes away with SELinux disabled. No AVCs in in the audit log either. The tabrmd is running in the `unconfined_service_t` domain. The change causing the issue is the switch from using pipes to sockets so `unconfined_service_t` and my client domain must be allowed to use pipes but not sockets.

flihp:
Any suggestions @martinezjavier? I can work around this easy enough locally but would be interested in your thoughts on a more permanent fix. Instructing people to put SELinux into permissive mode isn't an option. Creating a type for the tabrmd is probably the right thing to do here.

tstruk:
sorry, I didn't realize the '#' at the beginning of a line makes it look like this.

flihp:
well you found the github caps lock :smile: 
I'm gonna take a swing at implementing a unique SELinux domain for the daemon. We're gonna need it anyways. It'll have to go upstream to the refpolicy eventually.

martinezjavier:
@tstruk  I don't think your suggestion will work, since the AVC denial is not for the tpm2-abrmd process but for for the dbus-daemon:

```
type=AVC msg=audit(1507198100.003:2076): avc:  denied  { read write } for  pid=1076 comm="dbus-daemon" path="socket:[217760]" dev="sockfs" ino=217760 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=unix_stream_socket permissive=0
```

In fact, since the tpm2-abrmd is in an unconfined domain, the SELinux rules don't apply to it.

My understanding is that for the system bus, the dbus-daemon isn't allowed to read/write sockets created by processes in an unconfined domain. It works on the session bus because dbus-daemon also runs on an unconfined domain in that case:

```
$ ps -efZ | grep dbus-daemon
[...]

PR #205, opened 2017-10-05
SELinux: Add policy module for tabrmd.

flihp:
This is necessary because the default Fedora policy for the
unconfined_service_t type (the default context for a service w/o a
private type) doesn't allow the system_dbusd_t type to read / write
from it's stream sockets. Pipes were fine, but not stream sockets.

The policy module added here is a work around. The dbus policy module
doesn't have an interface to allow this type of access and so I've
hacked up this module to allow the operation. The form of the module
isn't right though and it's likely that we'll need to add such an
interface to the dbus module. This will require some collaboration /
vetting with the upstream refpolicy maintainers though.

The build system has not been updated to support building the policy
module so those on Fedora will have to build it manually. This isn't
hard though:
1) sudo dnf install selinux-policy-devel
2) pushd contrib && make -f /usr/share/selinux/devel/Makefile tabrmd.pp
3) sudo semodule -i tabrmd.pp
4) sudo restorecon /path/to/tpm2-abrmd
5) popd

At this point the new module should be loaded. You can verify this with
the semodule command. Additionally you should check to be sure the lable
on the label on the tpm2-abrmd binary is the `tabrmd_exec_t` type. Then
just stop & restart the daemon. It should now be running in the tabrmd_t
domain. You can verify this with the `ps` command.

`ps auxZ | grep tpm2-abrmd` should display something like:
system_u:system_r:tabrmd_t:s0   tss       6678  0.5  0.0 587400  4112 ?        Ssl  14:47   0:00 /usr/local/sbin/tpm2-abrmd

### DIFF
--- a/contrib/tabrmd.fc
+++ b/contrib/tabrmd.fc
@@ -1,0 +1,2 @@
+/usr/sbin/tpm2-abrmd	--	gen_context(system_u:object_r:tabrmd_exec_t,s0)
+/usr/local/sbin/tpm2-abrmd	--	gen_context(system_u:object_r:tabrmd_exec_t,s0)
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,12 @@
 .deps/
 .dirstamp
 .libs/
+tmp/
 *.la
 *.lo
 *.log
 *.o
+*.pp
 *.swp
 *.trs
 Makefile
--- a/contrib/tabrmd.if
+++ b/contrib/tabrmd.if
@@ -1,0 +1,1 @@
+## <summary></summary>
--- a/contrib/tabrmd.te
+++ b/contrib/tabrmd.te
@@ -1,0 +1,26 @@
+policy_module(tabrmd, 0.0.1)
+
+########################################
+#
+# Declarations
+#
+
+type tabrmd_t;
+type tabrmd_exec_t;
+init_daemon_domain(tabrmd_t, tabrmd_exec_t)
+
+allow tabrmd_t self:unix_dgram_socket { create_socket_perms };
+
+dev_rw_tpm(tabrmd_t)
+logging_send_syslog_msg(tabrmd_t)
+
+optional_policy(`
+    dbus_system_domain(tabrmd_t, tabrmd_exec_t)
+')
+
+# This next bit doesn't belong here. It should be exposed through an
+# interface likely from the dbus policy module.
+gen_require(`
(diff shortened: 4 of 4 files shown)


flihp:
this is a work around for #205

martinezjavier:
> this is a work around for #205

@flihp a typo, it's a workaround for #203.

I've tested your SELinux policy and it worked correctly for me:

```
$ ls -Z /usr/local/sbin/tpm2-abrmd                         
system_u:object_r:tabrmd_exec_t:s0 /usr/local/sbin/tpm2-abrmd

$ ps -efZ | grep abrmd
system_u:system_r:tabrmd_t:s0   tss      24194     1  0 01:05 ?        00:00:00 /usr/local/sbin/tpm2-abrmd

$ tpm2_listpcrs -L 0x4:7 -T tabrmd

** (process:24260): WARNING **: Failed to set fd 7 to NONBLOCK: Invalid argument

Bank/Algorithm: TPM_ALG_SHA1(0x0004)
PCR_07: 6d 72 06 87 1c 9c 6f 38 ad 39 97 ba ce eb ee 95 da de c0 4d
```
Do you also see that warning by the way?

contrib/tabrmd.fc
```diff
@@ -0,0 +1,2 @@
+/usr/sbin/tpm2-abrmd	--	gen_context(system_u:object_r:tabrmd_exec_t,s0)
+/usr/local/sbin/tpm2-abrmd	--	gen_context(system_u:object_r:tabrmd_exec_t,s0)
```
martinezjavier: Maybe in your commit message you can comment that the policy is for these two paths and if someone uses a different prefix, then the policy has to be modified.

In fact, I think it would be better to have the same information that's in the commit message in the README.

martinezjavier:
@flihp although it would be ideal to have this policy in the [refpolicy](https://github.com/TresysTechnology/refpolicy) repository, for Fedora is not strictly necessary. 

The Fedora selinux-policy package has patches on top of refpolicy for downstream policies, these are in the [selinux-policy](https://github.com/fedora-selinux/selinux-policy) and [selinux-policy-contrib](https://github.com/fedora-selinux/selinux-policy-contrib) repos.

So I think it would be great if the policy can be proposed there, that way users will just need to build from source or install the distro package and things would just work.

If you are not interested, I could attempt to propose a change to selinux-policy-contrib using your policy as a base.

PR #218, opened 2017-10-12
Add --flush-all command line option.

flihp:
This option causes the daemon to flush all transient objects, loaded
and active sessions from the TPM on startup. The purpose of this option
is to put the TPM into as much of a pristine state as we reasonably
can. We only flush tese particular handle type since they're the only
objects managed by the tabrmd.


flihp:
This should fail till https://github.com/01org/tpm2-tss/pull/573 is merged

PR #223, opened 2017-10-23
[master] test: remove all contexts after each testcase completes

jiazhang0:
It is necessary to clean up the unused contexts after a testcase
complete with a real TPM hardware, because a real TPM hardware always
has the limited resources, especially the RAM, to keep the transient
entries within it.

Note that the cleanup operation has to be executed in a clean sapi and
tcti contexts, because certain testcase may corrupt the state of tcti
context, causing 0xa0007 error code returned.


flihp:
Failed tests.

PR #226, opened 2017-10-24
[1.x] test: remove all contexts after each testcase completes

jiazhang0:
It is necessary to clean up the unused contexts after a testcase
complete with a real TPM hardware, because a real TPM hardware always
has the limited resources, especially the RAM, to keep the transient
entries within it.

Note that the cleanup operation has to be executed in a clean sapi and
tcti contexts, because certain testcase may corrupt the state of tcti
context, causing 0xa0007 error code returned.

### DIFF
--- a/test/integration/common.c
+++ b/test/integration/common.c
@@ -55,7 +55,7 @@ tcti_context_init (TSS2_TCTI_CONTEXT **tcti_context)
     g_debug ("context structure allocated successfully");
     rc = tss2_tcti_tabrmd_init (tmp_tcti_context, NULL);
     if (rc != TSS2_RC_SUCCESS)
-        g_error ("failed to initialize tcti context. RC: 0x%" PRIx32, rc);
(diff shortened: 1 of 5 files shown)


flihp:
Thanks for following up on this @jiazhang0. There are a few things I want to change but I've already held up this work for a bit while I was out of town so I'll do this after your work is merged.

jiazhang0:
@flihp Thanks. In addition, is it still necessary to create 2 PRs for master and 1.x when making a contribution each time? It looks 1.x is more like a development tree. If so,  I will only send the patch for 1.x and you will rebase it to the master branch.

flihp:
I should get this documented in the CONTRIBUTING.md file: My current workflow is to only commit changes to the master branch if they break backward compatibility. All other work (e.g. bug fixes, features that are backward compatible) into the 1.x release branch. Periodically I'm merging 1.x back into master so that everything ends up there but only backward compatible changes / features end up in the release branch.

PR #227, opened 2017-10-25
[master] test: support to run the test on a real TPM hardware

jiazhang0:
This change allows to run the integration test on a real TPM hardware
via specifying --enable-unit only. In addition, the end user must be
prompted with a highlight warning to understand the risk if continue.


flihp:
Closing: failed tests. This should be something we determine at configure time through an option as well. Something like `--enable-test-hwtpm` or something equally descriptive.

jiazhang0:
@flihp No problem. I will rework the patch after your latest commits are merged.

PR #228, opened 2017-10-25
[1.x] test: support to run the test on a real TPM hardware

jiazhang0:
This change allows to run the integration test on a real TPM hardware
via specifying --enable-unit only. In addition, the end user must be
prompted with a highlight warning to understand the risk if continue.


flihp:
see comment in #227

PR #235, opened 2017-11-02
Add guidelines for commit messages and branch selection

flihp:


CONTRIBUTING.md
```diff
@@ -21,6 +21,12 @@ consider this the window for comments.
 * All tests must pass on Travis CI for the merge to occur.
 * All changes must not introduce superfluous changes or whitespace errors.
 * All changes should adhere to the coding standard.
+* All commits should adhere to the git commit message guidlines described
+here: https://chris.beams.io/posts/git-commit/ with the following exceptions.
```
martinezjavier: Maybe it's better to refer to https://www.kernel.org/doc/html/latest/process/submitting-patches.html#the-canonical-patch-format which is a more formal and summarized version of what's mentioned in that article?

Or do you have any issues with the canonical patch format used by Linux?

CONTRIBUTING.md
```diff
@@ -21,6 +21,12 @@ consider this the window for comments.
...
+ * We allow commit subject lines less than 80 characters.
```
martinezjavier: Did you mean s/less/more ? Otherwise I'm not sure I follow this sentence.

flihp: haha yeah whoops, long day

CONTRIBUTING.md
```diff
@@ -21,6 +21,12 @@ consider this the window for comments.
...
+ * Commit subject lines should be prefixed with a string identifying the
+ effected subsystem / object. If the change is spread over a number of
+subsystems them the prefix may be omitted.
```
martinezjavier: Do you also want to require contributors to sign-off their commits? If so I would also mention here and add a reference to https://developercertificate.org/

CONTRIBUTING.md
flihp: I ran across both but the `canonical-patch-format` includes a lot more data than just commit message format and describes everything in the context of sending patches to a mailing list. I figured including all of that would probably confuse folks.

martinezjavier: @flihp fair enough.

CONTRIBUTING.md
flihp: that's a good idea

flihp:
Appreciate your feedback as always @martinezjavier.

Issue #239, opened 2017-11-09
GCC 5.4 complains -Werror=int-to-pointer-cast when


liuqun:
GCC 5.4 complains -Werror=int-to-pointer-cast when mocka test unit is enabled by "./configure --enable-unit"

https://github.com/intel/tpm2-abrmd/blob/bc1b31e7dab08f5d6151b0d72816fcce3d6eb022/test/command-attrs_unit.c#L114

https://github.com/intel/tpm2-abrmd/blob/bc1b31e7dab08f5d6151b0d72816fcce3d6eb022/test/command-attrs_unit.c#L145

```
$ ./bootstrap && ./configure --enable-unit && make check

...

test/command-attrs_unit.c: In function ‘__wrap_access_broker_lock_sapi’:
test/command-attrs_unit.c:114:12: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
     return (TSS2_SYS_CONTEXT*)mock ();
            ^
test/command-attrs_unit.c: In function ‘__wrap_Tss2_Sys_GetCapability’:
test/command-attrs_unit.c:145:21: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
     command_attrs = (TPMA_CC*)mock ();
                     ^
cc1: all warnings being treated as errors
```

```
gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/5/lto-wrapper
Target: arm-linux-gnueabihf
[...]

PR #241, opened 2017-11-09
test: Use built-in mock with cast functions instead of manual cast.

flihp:
This is intended to keep older versions of GCC happy.


flihp:
this should resolve #239

PR #246, opened 2017-11-15
autotools: distribute the template service file instead of the genera…

mgerstner:
…ted one

This fixes a bug in distribution tarballs created via `make dist`, that
caused the installed systemd service file to always contain paths
pointing to /usr/local/... instead of the actual prefixes used during
`configure` time.

PR #248, opened 2017-11-16
autotools: distribute the template service file instead of the genera…

flihp:
…ted one

This fixes a bug in distribution tarballs created via `make dist`, that
caused the installed systemd service file to always contain paths
pointing to /usr/local/... instead of the actual prefixes used during
`configure` time.


flihp:
This is the commit from #246 based on the 1.x release branch. Unfortunately changing the 'base' for the PR in #246 causes a pile of stuff to get pulled into the PR. Using this new PR as a work-around.

PR #249, opened 2017-11-16
autotools: distribute the template service file instead of the genera…

flihp:
…ted one

This fixes a bug in distribution tarballs created via `make dist`, that
caused the installed systemd service file to always contain paths
pointing to /usr/local/... instead of the actual prefixes used during
`configure` time.


flihp:
this PR must be merged by creating a merge commit

PR #252, opened 2017-11-24
SELinux cleanups

martinezjavier:
This pull requests renames the contrib directory to selinux, which is a more used convention in projects to store custom SELinux policy modules. I wonder though, if we shouldn't move to dist/selinux instead to keep all the dist files are there (probably adding subdirs for dbus, systemd, udev, etc).

It also adds a Makefile so users don't have to remember the commands to build and install the policy module.


flihp:
Just merged #253 which is what probably caused the build error here. I just pushed the rebuild button on travis-ci which should hopefully fix the build failure on this PR.

flihp:
@martinezjavier do you want this to go into the 1.x branch for the next release? I don't see any reason to leave it out given the delay in the tools and the fact that it's just a build change.

martinezjavier:
Yes please, I meant it be for both branches but keep getting confused about which project prefers the PR against 1.x or master.

flihp:
manually moved this over to 1.x and merged back into master.

Issue #254, opened 2017-11-27
man3/tss2_tcti_tabrmd_init_full.3 left around after uninstall


williamcroberts:
```
/workspace/tpm2-abrmd$ sudo make uninstall
 ( cd '/usr/local/etc/dbus-1/system.d' && rm -f tpm2-abrmd.conf )
 /bin/bash ./libtool   --mode=uninstall rm -f '/usr/local/lib/libtcti-tabrmd.la'
libtool: uninstall: rm -f /usr/local/lib/libtcti-tabrmd.la /usr/local/lib/libtcti-tabrmd.so.0.0.0 /usr/local/lib/libtcti-tabrmd.so.0 /usr/local/lib/libtcti-tabrmd.so /usr/local/lib/libtcti-tabrmd.a
 ( cd '/usr/local/include/tcti' && rm -f tcti-tabrmd.h )
 ( cd '/usr/local/share/man/man3' && rm -f tss2_tcti_tabrmd_init.3 )
 ( cd '/usr/local/share/man/man7' && rm -f tcti-tabrmd.7 )
 ( cd '/usr/local/share/man/man8' && rm -f tpm2-abrmd.8 )
 ( cd '/usr/local/lib/pkgconfig' && rm -f tcti-tabrmd.pc )
 ( cd '/usr/local/sbin' && rm -f tpm2-abrmd )
 ( cd '/usr/local/lib/systemd/system-preset' && rm -f tpm2-abrmd.preset )
 ( cd '/usr/local/lib/systemd/system' && rm -f tpm2-abrmd.service )
 ( cd '/usr/local/lib/udev/rules.d' && rm -f tpm-udev.rules )

$ cd /usr/local
$ find . | grep abrmd
./share/man/man3/tss2_tcti_tabrmd_init_full.3
```

flihp:
Good catch. I figured `distcheck` target would catch this (it does not). I guess the build will also leave the udev rules file around if a numeric prefix is provided. Fixing now ...

flihp:
should be resolved by #256 that just went into 1.x and master

PR #256, opened 2017-11-28
build: Add uninstall-local target to catch files missed by the defaul…

flihp:
…t rule.

Files left behind by the default rule are all special cases:
1) The tss2_tcti_tabrmd_init_full.3 man page is a soft link that we create
at install time.
2) If the udev rules file prefix is provided @ configure time then we
rename the file at install time.

The solution is the same for both: use the uninstall-local target.


flihp:
this resolves #254

Issue #262, opened 2017-12-09
code coverage


flihp:
The tss and tools already collect these metrics. Do the same here.

flihp:
Most of this has already gone in but I've run into a really weird edge case that's keeping us from enabling this in travis: It took some digging but gcov causes unit tests that mock the `open` system call to fail because for each test execution gcov tries to write a `.gcda` file and the mock function borks this up. The mock function will have to be able to identify the files opened by gcov and pass these through to the real `open` function. I'm looking into how to do this now.

flihp:
I just created #268 to track resolution of the mock / gcov issue. Its resolution will be sufficiently complex to warrant a separate fix and likely a work around / hack first.

flihp:
#270 is 90% of what we need and sufficient to satisfy this issue. I've created a milestone (to which this ticket belogs)  to track the lingering bits.

PR #263, opened 2017-12-09
1.x check code coverage

flihp:
incremental progress toward #262


flihp:
blocked on #264

flihp:
this is causing breakage in unit tests for some reason. Needs to be decomposed to find root cause

PR #264, opened 2017-12-11
build: Check that simulator has bound both ports.

flihp:
This is effectively port assignment collision detection. If the
simulator we start isn't bound to both ports we consider it an error. We
then kill the simulator and try again after an exponential backoff
period. If the simulator is able to bind to the data port but not the
command port it will still run but tests will fail. Killing it off is
important to keep from leaving tpm_server instances around.


flihp:
I've been putting off working on this for a while. Hunting down race conditions in the test harness isn't fun and it's not a new feature. Anyways this should fix the build failures I've been seeing in my work on #263. This is also very likely the same problem that @AndreasFuchsSIT worked around in the tss2 libs build: https://github.com/intel/tpm2-tss/commit/24fb1306e0204c931bb0cab63d04c6534d49bb78#diff-c949f93d03f44a4217d7a138f9e2e54a

flihp:
merging this more quickly than usual since it's blocking other work

PR #269, opened 2017-12-12
[master] test: support to run the test on a real TPM hardware

jiazhang0:
This change allows to run the integration tests on a real TPM hardware
via specifying --enable-test-hwtpm.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,6 +23,8 @@ and executed.
 * cmocka unit test framework
 * Microsoft / IBM Software TPM2 simulator version 532 as packaged by IBM:
 https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm532.tar
+* Alternately, run the test suite on a real TPM hardware, with a safety
+attention described below.
 
 # System User & Group
 As is common security practice we encourage *everyone* to run the `tpm2-abrmd`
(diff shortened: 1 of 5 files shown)


flihp:
Thanks for providing a PR against both branches. Since these changes are all backward compatible with the 1.x release branch I've reviewed the PR against 1.x. See: https://github.com/intel/tpm2-abrmd/blob/master/CONTRIBUTING.md#picking-a-branch

jiazhang0:
@flihp Note that  launching tpm2-abrmd in the patch for 1.x doesn't need --allow-root option but the patch for the master branch requires --allow-root option. I think they are not backward compatible each other.

flihp:
Sorry for the late reply @jiazhang0. This slipped through my filters.

The behavior you're seeing here is expected. The `--allow-root` option isn't "backward compatible" and so we've only added it to the master branch. If your system is configured such that the daemon must be run as root then you'll have to provide the `--allow-root` option on the command line. Worth noting: this is specifically intended to deter people from running the daemon as root. If you're accessing the device TCTI directly you should set up a udev rule to cause the `/dev/tpm*` device node to be owned by an unprivileged user instead of root.

jiazhang0:
@flihp Thanks for your comments about the background. However, this is the context of tpm hw testing, when launching `sudo make check`. The hw test needs to launch the newly built src/tpm2-abrmd which needs root permission to access /dev/tpm0 granted by the ownership tss:tss. This is why `sudo` is required. The user may manually run `sudo make install` next so we can assume current user may be the role of admin. So adding --allow-root in hw test script is reasonable for this context if using master branch. Currently, the use of `sudo make check` is working well with 1.x (of course it doesn't need --allow-root). Hopefully it also works with master branch.

flihp:
I think I understand now. So we need to add `--allow-root` to the test script in the master branch then?

jiazhang0:
@flihp Yes. I have fixed it https://github.com/tpm2-software/tpm2-abrmd/pull/308.

PR #271, opened 2017-12-12
1.x check code coverage

flihp:


flihp:
that's cute, thanks coveralls

PR #280, opened 2018-01-09
travis: Install liburiparser-dev.

flihp:
This is required by the upstream tpm2-tss.


flihp:
this will probably fail on account of a failing unit test. This is a new failure that I'm not able to reproduce locally. In the past similar failures have been caused by weirdness in the travis-ci build environment. There will be a ticket for this failure as soon as it's triggered here.

Issue #281, opened 2018-01-09
tss2-tcti-tabrmd_unit failing after upstream TCTI changes


flihp:
No clue what's causing this since it's only reproducible on travis-ci now. Will take some digging. Unfortunately the failure is in a file with ~30 unit tests and > 1000 LOC :frowning_face:

flihp:
Failure is in function: tcti_tabrmd_init_success_test. Unfortunately this looks like the unit test is attempting to acquire a connection with the dbus system bus which should only work on systems with a system bus. Since the travis build has never had a system bus available I'm not sure how this worked in the first place.

Switching this over to use the session bus may be a good short term fix. Implementing a proper mock object, or maybe just mocking the function (not sure which) that makes the connection will be sufficient.

flihp:
resolved in #283

liuqun:
Hi @flihp !
tcti_tabrmd_init_success_test() also crash when I try to run "test/tss2-tcti-tabrmd_unit".
It reports:
> ERROR: failed to allocate dbus proxy object: Cannot autolaunch D-Bus without X11 $DISPLAY

see bug report:
https://github.com/intel/tpm2-abrmd/issues/293

flihp:
Hey @liuqun. I replied to #293 with relevant data earlier today. All of the TCTI unit tests will produce the same problem and https://github.com/intel/tpm2-abrmd/issues/303 should cover the resolution.

PR #282, opened 2018-01-09
tests: stress test session save/load

williamcroberts:
At the 5th connection that saves and restores a session handle, abrmd
starts flushing handles and thus they are not available after a
client disconection.

This can be seen in the tools side as well with this simple
stress test:
  - https://github.com/williamcroberts/tpm2.0-tools/blob/session-tools/stress.sh

Changing this test, to try it 5 times causes the same failure, at 4
it works.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -80,6 +80,7 @@ TESTS_INTEGRATION = $(tests_integration)
 endif
 
 XFAIL_TESTS = \
+    test/integration/session-load-from-closed-connection.int \
     test/integration/start-auth-session.int \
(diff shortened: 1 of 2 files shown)


flihp:
this is gonna fail till I get a fix for #281 merged

flihp:
just merged #281. If you can rebase this it will now fail because your test is designed to hit a known bug. Might be worth adding this test to XFAIL when you rebase since it's expected to fail.

williamcroberts:
@flihp What known bug is it? I'll add it to XFAIL tomorrow.

PR #283, opened 2018-01-09
test: Connect tcti-tabrmd unit test to session bus instead of system.

flihp:
The travis-ci build magically started failing for this test. The root
cause is the initialization of the TCTI fails since the travis-ci
environment doesn't have a dbus system bus to connect to. It never has
though so I'm not sure why it started failing now.

This is a work around to use the session bus instead. We know a session
bus is available since we're the ones who set it up in the .travis.yml
file. It's likely better to create a proper mock object for the proxy
object, or maybe just figure out which function makes the connection and
mock that away.


flihp:
this resolves #281

PR #284, opened 2018-01-10
build: Include SELinux policy module in the release tarballs

martinezjavier:
The tarballs generated by the make dist target don't include the SELinux
policy module, so add it to the EXTRA_DIST variable.


martinezjavier:
@flihp right, sorry for missing this detail.

I'll respin the branch. Thanks for the feedback!

Issue #277, opened 2018-01-09
Update TCTI code to use new init functions


flihp:
see https://github.com/intel/tpm2-tss/pull/708

flihp:
#290 wrongly identified this issue as closed. That PR only covered the common case but ignored the configuration string we need to provide to the init function.

flihp:
The configuration string needs to allow the caller to specify the bus name / type in the same way the current `tss2_tcti_tabrmd_init_full` does. Since dbus hasn't standardized a method for specifying dbus connection data as a URI (though it has been discussed: https://lists.freedesktop.org/archives/dbus/2008-July/010019.html) we're left to our own devices.

flihp:
So really we won't be supporting a bus address as passed by the caller but instead allowing the caller to specify a string representation of one of two [GDbusType](https://developer.gnome.org/gio/stable/GDBusConnection.html#GBusType) enumerations. That's good enough for me currently.

PR #290, opened 2018-01-14
tcti: add spport for dynamic init interface

williamcroberts:
Add the dynamic tcti interface, and export the
module information and symbols.

Fixes: #277

### DIFF
--- a/src/tcti-tabrmd.c
+++ b/src/tcti-tabrmd.c
@@ -522,3 +522,26 @@ tss2_tcti_tabrmd_init (TSS2_TCTI_CONTEXT *context,
                                        TCTI_TABRMD_DBUS_TYPE_DEFAULT,
                                        TCTI_TABRMD_DBUS_NAME_DEFAULT);
 }
+
+static TSS2_RC
+tss2_tcti_tabrmd_init_config (TSS2_TCTI_CONTEXT *context,
+                              size_t            *size,
+                              const char        *conf)
+{
+	(void)conf;
+	return tss2_tcti_tabrmd_init(context, size);
+}
+
+/* public info structure */
+const static TSS2_TCTI_INFO tss2_tcti_info = {
+    .name = "tcti-abrmd",
+    .description = "TCTI module for communication with tabrmd.",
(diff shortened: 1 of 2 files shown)


src/tcti-tabrmd.c
```diff
@@ -522,3 +522,25 @@ tss2_tcti_tabrmd_init (TSS2_TCTI_CONTEXT *context,
...
+tss2_tcti_tabrmd_init_config (TSS2_TCTI_CONTEXT *context,
+                       size_t            *size,
+					   const char *conf) {
+	(void)conf;
+	return tss2_tcti_tabrmd_init(context, size);
+}
+
+/* public info structure */
+const static TSS2_TCTI_INFO tss2_tcti_info = {
+    .name = "tcti-socket",
```
williamcroberts: dang, so close....

src/tcti-tabrmd.c
```diff
@@ -522,3 +522,25 @@ tss2_tcti_tabrmd_init (TSS2_TCTI_CONTEXT *context,
                                        TCTI_TABRMD_DBUS_TYPE_DEFAULT,
                                        TCTI_TABRMD_DBUS_NAME_DEFAULT);
 }
+
+static TSS2_RC
+tss2_tcti_tabrmd_init_config (TSS2_TCTI_CONTEXT *context,
+                              size_t            *size,
+                              const char        *conf) {
```
williamcroberts: I could use G_GNUC_UNUSED from glib, not sure what your preference is: https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-GNUC-UNUSED:CAPS

(void) is compiler independent.

flihp:
This is great but it's only a partial implementation. The remaining work required is to define the configuration string so we can remove the old non-standard interface that allows the caller to specify the bus / name combo. I've been thinking about that a bit and will update #277 with details. Till then I'm gonna merge this to get the ball rolling.

Thanks again!

williamcroberts:
@flihp I didn't know that was part of it, I thought this was a no-argument TCTI. I can look into that more

PR #295, opened 2018-01-19
tcti-tabrmd: fix NULL conf string

jiazhang0:
The caller of tss2_tcti_tabrmd_init_config() may pass a NULL conf
string, e.g, launching tpm2_clear without any options, causing a
segment fault.

Signed-off-by: Jia Zhang <zhang.jia@linux.alibaba.com>

### DIFF
--- a/src/tcti-tabrmd.c
+++ b/src/tcti-tabrmd.c
@@ -632,6 +632,9 @@ tss2_tcti_tabrmd_init_config (TSS2_TCTI_CONTEXT *context,
         TCTI_TABRMD_DBUS_NAME_DEFAULT,
     };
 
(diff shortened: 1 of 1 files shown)


williamcroberts:
Yep seeing this pop up in the tools code: https://travis-ci.org/intel/tpm2-tools/jobs/330657610

williamcroberts:
Oh man I can't believe I approved this PR. On NULL we wanted to use the default bus configuration not return an error, ie NULL or "" should result in a default config, but "" goes through the parsing logic.

https://github.com/intel/tpm2-abrmd/pull/298

flihp:
happens to the best of us @williamcroberts :man_shrugging:

PR #297, opened 2018-01-19
Rename init_config function to align with naming from upstream.

flihp:
The general form in the spec will be Tss2_Tcti_<name>_Init where 'name'
is the name of the TCTI module.


flihp:
looks like this also fixes a whitespace error

PR #300, opened 2018-01-23
tcti: Migrate TCTIs to use new init interface.

flihp:
The old Init*Tcti functions from the upstream TCTIs have been deprecated
and replace with a standard Tss2_Tcti_<name>_Init function.


flihp:
merging this quick to fix build breakage caused by upstream changes

Issue #301, opened 2018-01-23
GCC 5.4(32-bits) complains when using "%lu" with a variable which is not "long unsigned int"


liuqun:
https://github.com/intel/tpm2-abrmd/blob/333824b5be1b5d39522491cfe38829ada01f70b3/src/tcti-tabrmd.c#L541-L546

```
In file included from /usr/include/glib-2.0/glib.h:62:0,
                 from /usr/include/glib-2.0/gobject/gbinding.h:28,
                 from /usr/include/glib-2.0/glib-object.h:23,
                 from /usr/include/glib-2.0/gio/gioenums.h:28,
                 from /usr/include/glib-2.0/gio/giotypes.h:28,
                 from /usr/include/glib-2.0/gio/gio.h:26,
                 from src/tcti-tabrmd.c:29:
src/tcti-tabrmd.c: In function ‘tabrmd_bus_type_from_str’:
src/tcti-tabrmd.c:546:14: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 4 has type ‘unsigned int’ [-Werror=format=]
     g_debug ("BUS_NAME_TYPE_MAP_LENGTH: %lu", BUS_NAME_TYPE_MAP_LENGTH);
              ^
/usr/include/glib-2.0/glib/gmessages.h:170:32: note: in definition of macro ‘g_debug’
                                __VA_ARGS__)
                                ^
cc1: all warnings being treated as errors
```

I guess that **size_t** was defined as **unsigned int** on 32-bit architecture(arm and x86), but defined as **long unsigned int** on x86_64 architecture.

We may cast BUS_NAME_TYPE_MAP_LENGTH from `size_t` into `long unsigned int`:
```
#define BUS_NAME_TYPE_MAP_LENGTH \
    ((long unsigned int) (sizeof (bus_name_type_map) / sizeof (bus_name_type_entry_t)))

void foobar()
{
    g_debug ("BUS_NAME_TYPE_MAP_LENGTH: %lu", BUS_NAME_TYPE_MAP_LENGTH);
}
```

Casting BUS_NAME_TYPE_MAP_LENGTH into **int** also works for me:
```
#define BUS_NAME_TYPE_MAP_LENGTH \
    ((int) (sizeof (bus_name_type_map) / sizeof (bus_name_type_entry_t)))

void foobar()
{
[...]

PR #304, opened 2018-01-23
tcti-tabrmd: Fix format specifier in debug statement.

flihp:
size_t divided by size_t equals size_t. Not sure why I picked a long
unsigned int for the format specifier in the first place.


flihp:
this resolves #301

PR #306, opened 2018-01-23
docs: Use coveralls badge from shields.io instead of coveralls.

flihp:
According to https://github.com/lemurheavy/coveralls-public/issues/116
the github web caching is aggressive to the point of causing the badge
URL provided by coveralls to show old cached values for a very long
time. A number of hacks were proposed but the most sane seemed to be to
use the badge provided by shields.io instead as it doesn't suffer from
the same issue as the one from coveralls.


williamcroberts:
Awesome I have had the same issue in tools, was wondering what to do about it.

PR #311, opened 2018-02-02
Fix the inappropriate use of g_error with GError.

lugong:
This commit is try to fix this issue https://github.com/tpm2-software/tpm2-abrmd/issues/96 .


lugong:
@flihp, thanks for your review.

I will use the 1.x branch to create a new pull request.

Issue #96, opened 2017-06-29
tcti-tabrmd abuses g_error


flihp:
See: https://github.com/01org/tpm2-abrmd/blob/master/src/tcti-tabrmd.c#L475

`g_error` causes the application to abort. This should only be used for errors that can't be handled with a return value. 6b5ca2c7f82df7ef97a2f40d23829962f02e7a7e is a good example of the right way to do this for anyone who wants to help fix this.

lugong:
Hi @flihp,
I have created a pull request to fix this issue. [https://github.com/tpm2-software/tpm2-abrmd/pull/311](url)
Please help to review this modification.

flihp:
Just added this to the 2.0 milestone. Must be resolved before 2.0 release.

PR #315, opened 2018-02-05
Fix the inappropriate use of g_error with GError.

lugong:
This commit is try to fix this issue #96 .

Signed-off-by: Lu Gong <lu.gong@intel.com>


flihp:
Hey @lugong. Looks like you've included a commit with your work on the TCP/IP/TLS frontend. Was this intentional?

lugong:
No. The code related to IPC TLS is by accident.
I though the pull request would not be affected if I push a new commit to the branch, but It seems the pull request will update automatically.
I have removed the new commitment and the pull request recovered.

PR #319, opened 2018-02-07
tcti-dynamic: Add support for dynamic loading of TCTIs and standard i…

flihp:
…nit.

This integrates a new Tcti object into the daemon. The TctiFactory knows
it as "dynamic". it's selected by providing `--tcti=dynamic` on the
command line. Two more options are provided to allow for selection of the
TCTI library (same search rules as `dlopen`) and a configuration string.
To use this interface the TCTI must support the new initialization mechanism
currently being standardized.

The tabrmd manpage is also updated for the new options associated with the
new TCTI loading / initialization mechanism.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,6 +33,7 @@ TESTS_UNIT = \
     test/random_unit \
     test/session-entry_unit \
     test/test-skeleton_unit \
+    test/tcti-dynamic_unit \
     test/tcti-echo_unit \
     test/tcti-factory_unit \
     test/thread_unit \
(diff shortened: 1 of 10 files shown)


williamcroberts:
Why not just do something easier...

--tcti=<name>:<config-string>

Try what they gave raw with dlopen.
Try prefixing with tcti-
See: https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_tcti_ldr.c#L67

Then you don't need to add this dynamic option.

Issue #323, opened 2018-02-08
You can't dlclose until the END of pointer usage


williamcroberts:
The gist, you can't hand back pointers for use and then call dlclose().

In the code tcti_dynamic_discover_info() the tcti is dlopened and the info struct
is set in the self pointer:
```
    self->tcti_info = info_func ();
```
At the end of tcti_dynamic_discover_info() you call dlclose() invalidating the self->tcti_info
pointer as dlclose un-maps it from the processes space.

You need to dlclose at the exit point of the daemon. You can look at the tpm2-tools code [here](https://github.com/tpm2-software/tpm2-tools/commit/4019681908487bb26829f564d1384857de1ce012#diff-970b42817994724c3f92f8e987d16029R175)

The issue I was working around with ```--disable-dlclose``` is a bug described in the tpm2-tools commit message [here](https://github.com/tpm2-software/tpm2-tools/commit/130582559d7c51d18e3ce82803c30bc161d9c34d#diff-9a57bc4e2757e72fe020a15ac7775dfa)

You probably were not hitting this because you had ```--disable-dlclose``` (presumably) in your configure script.

flihp:
I was probably tripping over this bug and thought it was the issue with dlclose in the travis-ci image. I put the --disable-dlclose on the configure line and just left it there for all of my local builds which caused them to effectively disable this bug. Thanks for helping with the debugging. I'll knock out a patch.

williamcroberts:
thank @kaccardi who wasted more of her life today :1st_place_medal:

flihp:
Wasn't wasted time if she found my bug. Thanks @kaccardi!

flihp:
should be fixed in #324

PR #324, opened 2018-02-09
tcti-dynamic: dlclose tcti libraries when the TCTI is no longer in use.

flihp:
This resolves a bug in the current implementation. Previously we were
dlclosing the TCTI library after tcti context intitialization. This
caused a segfault on subsquent use of the TCTI context since the library
had been dlclosed / unloaded.

This patch ties the lifetime of the library handle (handle returned by
dlopen) to the lifetime of the TctiDynamic object.


flihp:
this resolves #323

PR #328, opened 2018-02-14
Dynamic tcti default

### LINKED ISSUE #327: sane default TCTI
It looks like, when provided with no `--tcti` option we're not picking a sane default value for the TCTI library. If the option is omitted we should default to the device TCTI.

flihp:
this resolves #327


flihp:
clang so picky

PR #331, opened 2018-02-16
D-Bus: install a service file to allow on-demand systemd unit activation

mgerstner:
Maybe you'd like to add these bits. A D-Bus service file allows to start the daemon on demand without having to explicitly enable the system unit.


flihp:
Perfect timing @mgerstner, I'll make sure this gets into the next 1.x release. Thanks!

PR #332, opened 2018-02-16
ci: Install libgcrypt development package.

flihp:
This is a new dependency for the ESAPI library that's now in the
tpm2-tss repo.


flihp:
pushing this through quickly since it fixes build breakage caused by a new dependency introduced upstream

PR #343, opened 2018-02-23
Clarify UID restriction to users tss and root. Fixes #342

### LINKED ISSUE #342: Clarify error messages and doc about running tpm2-abrmd as users tss and root
Clarify error messages and doc about running tpm2-abrmd as users tss and root.
The current messages and INSTALL.md doc should mention that tpm2-abrmd needs to run as user tss or root (non-root is preferred).

$ src/tpm2-abrmd --tcti=socket
Failed to acquire DBus name com.intel.tss2.Tabrmd. UID 1000 must be allowed to "own" this name. Check DBus config.

# src/tpm2-abrmd
refusing to run as root. pass --allow-root if you know what you're doing. #
 [MISSING new line so shell prompt appears on same line]

dantpm:
Clarified error messages in tpm2-abrmd when
* not ran as user tss or root
* ran as root without --with-root

Testing
Ran as user tss, and as root with tpm2-abrmd --with-root
Negative tests: ran as root (without --with-root) and as non-root/non-tss user.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,6 +27,7 @@ https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm532.tar
 attention described below.
 
 # System User & Group
+`tpm2-abrmd` must run as user `tss` or `root`.
 As is common security practice we encourage *everyone* to run the `tpm2-abrmd`
 as an unprivileged user. This requires creating a user account and group to
 use for this purpose. Our current configuration assumes that the name for this
@@ -43,9 +44,12 @@ $ sudo useradd --system --user-group tss
 You may wish to further restrict this user account based on your needs. This
 topic however is beyond the scope of this document.
 
+To run tpm2-abrmd as root, which is not recommended, use the `--allow-root`
+option.
+
 # Obtaining the Source Code
 As is always the case, you should check for packages available through your
-Linux distro before you attepmt to download and build the tpm2-abrmd from
+Linux distro before you attempt to download and build the tpm2-abrmd from
 source code directly. If you need a newer version than provided by your
 Distro of choice then you should download our latest stable release here:
 https://github.com/01org/tpm2-abrmd/releases/latest.
@@ -96,7 +100,7 @@ all supported options.
 The `tpm2-abrmd` claims a name on the D-Bus system bus. This requires policy
 to allow the `tss` user account to claim this name. By default the build
 installs this configuration file to `${sysconfdir}/dbus-1/system.d`. We allow
-this to be overriden using the `--with-dbuspolicydir` option.
+this to be overridden using the `--with-dbuspolicydir` option.
 
 Using Debian (and it's various derivatives) as an example we can instruct the
 build to install the dbus policy configuration in the right location with the
--- a/src/ipc-frontend-dbus.c
+++ b/src/ipc-frontend-dbus.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, Intel Corporation
+ * Copyright (c) 2017 - 2018, Intel Corporation
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -276,7 +276,7 @@ get_pid_from_dbus_invocation (GDBusProxy            *proxy,
     }
 }
 /*
- * Generate a random uint64 returned in the id out paramter.
+ * Generate a random uint64 returned in the id out parameter.
  * Mix this random ID with the PID from the caller. This is obtained
  * through the invocation parameter. Mix the two together using xor and
  * return the result through the id_pid_mix out parameter.
@@ -439,7 +439,7 @@ on_handle_create_connection (TctiTabrmd            *skeleton,
  * requesting that an outstanding TPM command should be canceled. It is
  * registered with the Tabrmd in response to acquiring a name
  * on the dbus (on_name_acquired). It does X things:
- * - Locate the Connection object associted with the 'id' parameter in
+ * - Locate the Connection object associated with the 'id' parameter in
  *   the ConnectionManager.
  * - If the connection has a command being processed by the tabrmd then it's
  *   removed from the processing queue.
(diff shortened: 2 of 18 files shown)


dantpm:
Fixes bug #342.

INSTALL.md
```diff
@@ -27,6 +27,7 @@ https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm532.tar
 attention described below.
 
 # System User & Group
+`tpm2-abrmd` must run as user `tss` or `root`.
```
flihp: This isn't strictly true is it? I think I see what you're getting at: the default is the user 'tss' but you can really run it as any user so long as the dbus config is right. I think the text following this addition says exactly this?

INSTALL.md
```diff
@@ -43,9 +44,12 @@ $ sudo useradd --system --user-group tss
 You may wish to further restrict this user account based on your needs. This
 topic however is beyond the scope of this document.
 
+To run tpm2-abrmd as root, which is not recommended, use the `--allow-root`
+option.
```
flihp: +1 - Thanks for this. Should have been in here already.

INSTALL.md
dantpm: Yes it is true, unless you hand-edit the undocumented tpm2-abrmd.conf and add some XML to allow another user.
Or are you referring to "sudo"? Which does change the user to root.

I can add text about hand-editing tpm2-abrmd.conf as an alternative.

martinezjavier: > This isn't strictly true is it? I think I see what you're getting at: the default is the user 'tss' but you can really run it as any user so long as the dbus config is right. I think the text following this addition says exactly this?

This is only partially true though, because is not only the access to the system bus, the daemon should also have access to the TPM2 device. And currently the user and group are hardcoded to `tss`.

martinezjavier:
@dantpm I think your changes are good if you split the typo fixes in a different commit as suggested by @flihp and have a proper commit format as mentioned in the other PR.

As mentioned, I believe what you said is true due the udev rule we currently ship. We can decide later if we want to change that (i.e: providing a configure option to choose a different default user) but for now I think the docs should honor the current constraints.

dantpm:
1. I will add a Signed-off-by: tag.

2. But the commit message is already there.  I added it with:
**git commit -m "Add sanity checks to int-log-compiler.sh. Fixes #764"**
I even see it on this webpage as the header and it is also displayed under the "commit" tag.
But perhaps the real problem is that the commit message is in the wrong format. If that is the case, what is the issue? The message is <80 characters and has the bug#, but maybe something else is needed?  Or perhaps there is another place where I need to add the (already existing) commit message?

3. I can update and merge to the branch tied to this pull request, but I see multiple merge and commit messages which look messy.  We called them "merge turds" in Solaris and was not allowed. That's why I created a fresh pull request (I recreated the branch) so remove the multiple merge and commit messages.

martinezjavier:
> But the commit message is already there. I added it with:
> git commit -m "Add sanity checks to int-log-compiler.sh. Fixes #764"

Yes, but that's exactly what I was pointing out. If you look at the [documented](https://chris.beams.io/posts/git-commit/) referred in the CONTRIBUTING document, you will see what's not correct in your commit message. The summary of that blog post is:

```
The seven rules of a great Git commit message:

1. Separate subject from body with a blank line
2. Limit the subject line to 50 characters
3. Capitalize the subject line
4. Do not end the subject line with a period
5. Use the imperative mood in the subject line
6. Wrap the body at 72 characters
7. Use the body to explain what and why vs. how
```

You are not following most of these rules, for example I would use the following commit message:

```
INSTALL.md: Clarify UID restriction to users tss and root

The tpm2-abrmd daemon can't be run as any user, there are some restrictions
on which users can run it. These restrictions are due two reasons:

1) Access to the D-Bus system bus is resctricted, so a D-Bus config file is
   shipped that only allows the root and tss users to aquire the tpm2-abrmd
   D-Bus well-known-name.
2) Access to the TPM2 character devices (/dev/tpm*) are only accessible by
   the tss and root users. A udev rule is shipped that changes ownership of
   these character device nodes.

Clarify these restrictions in the docs so users are aware of them. Also, by
being there due some typos cleanups.

Fixes #342

```

> I can update and merge to the branch tied to this pull request, but I see multiple merge and commit messages which look messy.

But you can avoid that by rewriting the history of your branch. That's OK for development branches, these can change while being proposed for merging. This is usually my workflow:

```
$ git checkout -b fix-foo-bar origin/master
$ $EDITOR file.c
$ git log --pretty=one --abbrev-commit --no-merges # I use this to know the correct subject line to use
$ git commit -s -a -m "file: Fix foo bar"
$ git commit --amend # I write a proper commit message here with Fix tags, etc
$ git push $myrepo fix-foo-bar

Open a pull request on Github, and if I get some feedback to change things I do:

$ git commit -a --fixup=$SHA1 # $SHA1 is the commit I want to fixup
$ git rebase -i $SHA1~1
$ git push $myrepo fix-foo-bar -f
```

Github will take care of refreshing the web UI so the patches for your new rebased branch are shown.

Something I don't like about the Github UI is that it doesn't show the diff between the current and previous version, so I usually edit the pull request comment section and add:

Changes since v1:
- Changed bar for baz as suggested by John Doe.

@dantpm I hope all this info is helpful to you!

INSTALL.md
dantpm: So this is completely true, because of the daemon and the /dev/tpm0 device.  I filed a separate bug to add documentation on adding users for abrmd and perhaps (I don't recommend) add a configure flag to change the default user from tss to some other user:
#349 Document how to add to permitted abrmd users and maybe how to change default user, tss

dantpm:
Added Signed-off-by line and more text to the commit block.

martinezjavier:
@dantpm did you forget to update your remote branch? Because I still see the old commit without S-o-B and message body.

dantpm:
No I did not forget...it is still in the queue. I spent all yesterday merging and it still does not work.
For example, I get old commits in my branch listed to be merged that are already in the upstream master branch.  Or it asks to resolve conflicts on files I never changed during merging. That's what I mean when I say "merge hell".

dantpm:
I split this pull request into 2 issues and 2 pull requests:
Issue #342 for *.c code changes only
Issue #356 for INSTALL.md changes only

Issue #344, opened 2018-02-23
build error


williamcroberts:
src/tcti-tabrmd.c: In function â€˜init_tcti_dataâ€™:
src/tcti-tabrmd.c:396:42: error: assignment from incompatible pointer type [-Werror]
     TSS2_TCTI_TRANSMIT (context)         = tss2_tcti_tabrmd_transmit;
                                          ^

williamcroberts:
I'll look into this

martinezjavier:
@williamcroberts already have a patch for that.

dantpm:
I just noticed the build error.  I just pulled new source today.

PR #345, opened 2018-02-23
tcti: fix build error due wrong TCTI transmit functions signatures

martinezjavier:
The third parameter of the TCTI transmit functions has a const qualifier
but this is missing in the tss2_tcti_{echo,tabrmd}_transmit functions.

This causes a build error due an incompatible pointer type assignment.

Fixes: #344


williamcroberts:
```
  CC       test/test_command_attrs_unit-command-attrs_unit.o
test/access-broker_unit.c: In function â€˜access_broker_setup_with_initâ€™:
test/access-broker_unit.c:155:64: error: assignment from incompatible pointer type [-Werror]
     TSS2_TCTI_TRANSMIT (tcti_peek_context (TCTI (data->tcti))) = __wrap_tcti_echo_transmit;
                                                                ^
cc1: all warnings being treated as errors
```

martinezjavier:
@williamcroberts thanks, I missed since that was compile out locally with my configure options. Just pushed a new version.

PR #350, opened 2018-02-27
update YES/NO definitions

williamcroberts:
tpm2-tss removed YES/NO macros, so replace them with their
original values.

### DIFF
--- a/src/resource-manager.c
+++ b/src/resource-manager.c
@@ -687,7 +687,7 @@ vhandle_iterator_callback (gpointer entry,
              PRIu32, state->max_count, cap_data->data.handles.count);
     /* if we've collected max_count handles set 'more_data' and return */
     if (!(cap_data->data.handles.count < state->max_count)) {
-        state->more_data = YES;
+        state->more_data = 1;
         return;
     }
     cap_data->data.handles.handle [cap_data->data.handles.count] = vhandle;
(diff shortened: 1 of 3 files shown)


williamcroberts:
@flihp I could also see:
1. #defining these in your header file somewhere...
2. Dropping the logic comparisons directly, ie `if (more_data)` not `if (more_data == 1)`

Let me know what you prefer here.

martinezjavier:
@williamcroberts can't we just change the more data type to `bool` from `stdbool` or `gboolean` from `glib` ? AFAICT this is only used to keep state that's internal to the RM and isn't something that's passed to the SAPI and TCTI layers.

williamcroberts:
@martinezjavier that sounds reasonable, lets see what @flihp wants

flihp:
sorry for the delay, working through this now

src/tcti-tabrmd.c
```diff
@@ -658,7 +658,7 @@ Tss2_Tcti_Tabrmd_Init (TSS2_TCTI_CONTEXT *context,
 const static TSS2_TCTI_INFO tss2_tcti_info = {
     .name = "tcti-abrmd",
     .description = "TCTI module for communication with tabrmd.",
-    .config_help = "This module takes NO arguments.",
+    .config_help = "This module takes 0 arguments.",
```
flihp: well, at least the comment is still accurate? 😄

williamcroberts:
At first I though gboolean was a good suggestion, but eventually it would propagate to build_cap_handles_response(), which requires that type (it's a byte).

Id prefer not to have to truncate a gboolean (int) when assigning to TPMI_YES_NO (unsigned char).

Ill post the patches and fix this one as well.

src/tcti-tabrmd.c
williamcroberts: Fixed

williamcroberts:
Boolean version here: https://github.com/tpm2-software/tpm2-abrmd/pull/351

PR #353, opened 2018-03-01
tcti: Allow tss2_tcti_tabrmd_finalize() to be called twice

liuqun:
tss2_tcti_tabrmd_finalize() might get called more than once from the client-side when their implementation is buggy.
Add a defensive checking to prevent client application from crashing.

### DIFF
--- a/src/tcti-tabrmd.c
+++ b/src/tcti-tabrmd.c
@@ -293,6 +293,10 @@ tss2_tcti_tabrmd_finalize (TSS2_TCTI_CONTEXT *context)
         g_warning ("Invalid parameter");
         return;
     }
+    if (TABRMD_STATE_FINAL == TSS2_TCTI_TABRMD_STATE (context)) {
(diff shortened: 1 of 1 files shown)


flihp:
Hi @liuqun. If you can rebase this on the current master the CI should pass. sourceforge is down again and the CI depneds on it. I've set up a mirror to get us through the outage (and likely future ones too).

flihp:
Based on this PR / bug report (crashing clients is a bug) I went ahead and created a test case to accompany the fix in this PR. I was pretty sure, based on the code and our use of `g_clear_object`, that there isn't a crash but I wanted to be sure. The test case passes after calling the finalize function twice so I don't think this patch is necessary. Can you confirm?

liuqun:
Thanks @flihp ! 
You're right. g_clear_object(GObject **) will release the resource then assign sock_connect with NULL immediately.
tss2_tcti_tabrmd_finalize() is always safe, even when it was called twice.

PR #354, opened 2018-03-01
ci: Use mirrored cmocka and simulator tarballs.

flihp:
sourceforge has been down on and off for a week. Cmocka.org had some
troubles last week too. Switching to a mirror to try to keep from having
any more down time in the future.


flihp:
merging this quick to work around yet another sourceforge outage

PR #363, opened 2018-03-05
autotools: Include tpm2-abrmd preset template in distribution tarball

martinezjavier:
A generated tpm2-abrmd.preset file is included in the distribution tarball, so
there's no way to change the default preset policy by using a configure option.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -139,7 +139,7 @@ EXTRA_DIST = \
     man/tpm2-abrmd.8.in \
     dist/tpm2-abrmd.conf \
     dist/tcti-tabrmd.pc.in \
-    dist/tpm2-abrmd.preset \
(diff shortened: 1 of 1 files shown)


flihp:
looks like we're gonna have a 1.3.1 pretty quickly :)

martinezjavier:
@flihp yeah, sorry for not catching this before. The thing is that I can't update the Fedora tpm2-abrmd package until the one that ships the SELinux module is accepted (still under review).

So I don't pay as much attention to tpm2-abrmd as I do to the tools and tss.

flihp:
np, i just fixed it up

flihp:
will have an RC0 for 1.3.1 once I merge this and sign the tag

flihp:
closing in favor of fixup in #371

PR #368, opened 2018-03-06
INSTALL.md: Clarify UID Restrictions to Users tss and root

### LINKED ISSUE #336: Document configure flags better
Add configure flags to override:
*  /usr/local/etc/default
     This flag will be --with-sysdefaultdir.  E.g., $ configure --with-sysdefaultdir=/etc/default

* /usr/local/share/dbus-1/system-services
     This flag will be --with-dbusdatadir.  E.g., 
     $ configure --with-dbusdatadir=/usr/share/dbus-1/system-services

Update INSTALL.md with the new flags.  Also add these clarifications:
* Change version of IBM TPM2 simulator from 532 to 974.
* Mention default values for GNU installation directories at:
https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
* Add missing example:
	--with-systemdpresetdir=/lib/systemd/system-preset
* Add missing dependency on D-Bus 1 libraries
* Typo: s/overriden/overridden/

### LINKED ISSUE #342: Clarify error messages and doc about running tpm2-abrmd as users tss and root

dantpm:
Point to the document for GNU installation directory defaults.

Add example for --with-systempresetdir option.

Document --datarootdir option (used for Tabrmd.service).

Document that the D-Bus 1 library and header files are required.

Fixes #336.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,6 +12,7 @@ required:
 * pkg-config
 * glib 2.0 library and development files
 * libsapi and TCTI libraries from https://github.com/01org/TPM2.0-TSS
+* D-Bus 1 library and header files
 
 **NOTE**: Different GNU/Linux distros package glib-2.0 differently and so
 additional packages may be required. The tabrmd requires the GObject and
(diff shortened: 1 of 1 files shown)


dantpm:
Removed spelling error fixes (already fixed in another bug).
Removed Microsoft/IBM simulator version update (in separate bug #379)

PR #371, opened 2018-03-07
autotools: Include tpm2-abrmd preset template in distribution tarball

flihp:
A generated tpm2-abrmd.preset file is included in the distribution tarball, so
there's no way to change the default preset policy by using a configure option.
This commit causes the build to distribute the template file so that
builds from source will generate the final preset file with data from
the configure script. Additionally, the definition of the
systemdpreset_DATA variable is now conditionally defined alongside the
other systemd stuff.


flihp:
This is a fixup for #363

PR #389, opened 2018-03-16
Remove use of tpm20.h and use the appropriate TSS2 headers.

flihp:
The usptream tss2 libs and headers have seen some significant cleanup
and the removal of the tpm20.h header was one of them. This commit goes
through just about every source file in this repo and replaces the use
of tpm20.h with the appropriate header from upstream.

There were a lot of places where we were relying on tpm20.h to include
string.h, stdlib.h etc for us and so we've had to add these to the
offending source file here.


flihp:
look @williamcroberts, it passed!

PR #390, opened 2018-03-16
AccessBroker: define ABI version locally

tstruk:
TSS doesn't export the ABI defines in public heades anymore.
This is a follow up for https://github.com/tpm2-software/tpm2-tss/pull/864

### DIFF
--- a/src/access-broker.c
+++ b/src/access-broker.c
@@ -156,6 +156,15 @@ access_broker_class_init (AccessBrokerClass *klass)
                                        N_PROPERTIES,
                                        obj_properties);
 }
+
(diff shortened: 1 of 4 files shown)


flihp:
this is blocked on #391

flihp:
This PR will fail till the duplicate symbols are removed upstream.

flihp:
IMHO we should probably used different names here anyways since we've been replacing `SAPI` with `sys` everywhere we can.

tstruk:
Do you have any good name in mind?

tstruk:
PR updated

PR #407, opened 2018-03-30
Added cc -Wextra Flag

### LINKED ISSUE #406: Add cc -Wextra flag for extra compile-time checking
Add cc -Wextra for extra compile-time source checking and fix or exclude warnings generated.

danintel:
Added cc -Wextra for extra compile-time source checking.
Fixed or excluded all new warnings.

Fixes #406.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -180,6 +180,8 @@ VALGRIND_memcheck_FLAGS = --leak-check=full --show-leak-kinds=definite,indirect 
 src/ipc-frontend-dbus.c : $(BUILT_SOURCES)
 
 # utility library with most of the code that makes up the daemon
+# -Wno-unused-parameter for automatically-generated tabrmd-generated.c:
+src_libutil_la_CFLAGS  = $(AM_CFLAGS) -Wno-unused-parameter
 src_libutil_la_LIBADD  = $(DBUS_LIBS) $(GIO_LIBS) $(GLIB_LIBS) $(PTHREAD_LIBS) \
     $(TSS2_SYS_LIBS) $(TCTI_DEVICE_LIBS) $(TCTI_SOCKET_LIBS)
 src_libutil_la_SOURCES = \
--- a/src/access-broker.c
+++ b/src/access-broker.c
@@ -102,7 +102,7 @@ access_broker_get_property (GObject     *object,
     }
 }
 /*
- * Dispose function: relese references to gobjects. We also free the SAPI
+ * Dispose function: release references to gobjects. We also free the SAPI
  * context here as well. Typically freeing this data would be done in the
  * finalize function but it can't be used w/o the underlying TCTI.
  */
@@ -231,7 +231,7 @@ access_broker_lock (AccessBroker *broker)
             g_error ("AccessBroker: attempted to lock uninitialized mutex");
             break;
         default:
-            g_error ("AccessBroker: unkonwn error attempting to lock SAPI "
+            g_error ("AccessBroker: unknown error attempting to lock SAPI "
                      "mutex: 0x%x", error);
             break;
         }
@@ -434,7 +434,7 @@ access_broker_get_response (AccessBroker *broker,
 /**
  * In the most simple case the caller will want to send just a single
  * command represented by a Tpm2Command object. The response is passed
- * back as the return value. The resonse code is returend through the
+ * back as the return value. The response code is returned through the
  * 'rc' out parameter.
  * The caller MUST NOT hold the lock when calling. This function will take
  * the lock for itself.
@@ -511,7 +511,7 @@ access_broker_new (Tcti *tcti)
  * Initialize the AccessBroker. This is all about initializing internal data
  * that normally we would want to do in a constructor. But since this
  * initialization requires reaching out to the TPM and could fail we don't
- * want to do it in the constructur / _new function. So we put it here in
+ * want to do it in the constructor / _new function. So we put it here in
  * an explicit _init function that must be executed after the object has
  * been instantiated.
  */
--- a/src/access-broker.c
+++ b/src/access-broker.c
@@ -37,6 +37,24 @@
 #include "tpm2-response.h"
 #include "util.h"
 
+#define TPMS_CAPABILITY_DATA_ZERO_INIT { 0, { 0, { \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, \
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 } } }
+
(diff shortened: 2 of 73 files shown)


dantpm:
No there's just spelling errors.  There are a lot of them and I will separate them out.

The only things hiding in the code is bugs we have yet to find.

dantpm:
I separated out the spelling error-only fixes into separate commit.
So there are two commits in this pull request.

src/access-broker.c
```diff
@@ -522,6 +522,7 @@ access_broker_init_tpm (AccessBroker *broker)
 out:
     return rc;
 }
+
```
flihp: no superfluous whitespace changes please: https://github.com/tpm2-software/tpm2-abrmd/blob/master/CONTRIBUTING.md#patch-requirements

src/access-broker.c
```diff
@@ -532,7 +533,23 @@ access_broker_get_trans_object_count (AccessBroker *broker,
...
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 } } };
```
flihp: This is pretty unwieldy and largely meaningless. I tried enabling `-Wextra` in the past and stuff like this is what got me to back away from the whole thing. Do we have any options to trim this stuff down?

src/access-broker.c
```diff
@@ -667,7 +685,23 @@ access_broker_flush_all_unlocked (AccessBroker     *broker,
...
```
flihp: Having this block in two separate tests is begging for at least a macro like the other structure initialization macros we have in the test harness. I think they all have the `STATIC_INIT` suffix.

src/random.c
```diff
@@ -144,10 +144,10 @@ random_get_bytes (Random    *random,
     }
     return i;
 }
+
```
flihp: whitespace: https://github.com/tpm2-software/tpm2-abrmd/blob/master/CONTRIBUTING.md#patch-requirements

src/random.c
```diff
@@ -144,10 +144,10 @@ random_get_bytes (Random    *random,
...
 /*
- * Get 64 bits of random data from the parameter 'random' object. If
- * we successfully fill in the supplied 'dest' parameter with random
- * data we return 0. On error return -1.
+ * Get 64 bits of random data from the parameter 'random' object.
+ * On error, return 0 instead of random data.
```
flihp: This change isn't mentioned anywhere in the commit message but it doesn't look related to new CFLAGS you've added.

src/random.c
```diff
@@ -160,10 +160,10 @@ random_get_uint64 (Random      *random)
 
     return dest;
 }
+
 /*
- * Get 32 bits of random data from the parameter 'random' object. If
- * we successfully fill in the supplied 'dest' parameter with random
- * data we return 0, on error return -1.
+ * Get 32 bits of random data from the parameter 'random' object.
+ * On error, return 0 instead of random data.
```
flihp: same as last comment

src/random.c
```diff
@@ -160,10 +160,10 @@ random_get_uint64 (Random      *random)
 
     return dest;
 }
+
```
flihp: whitespace

src/random.c
```diff
@@ -176,6 +176,7 @@ random_get_uint32 (Random       *random)
...
```
flihp: whitespace

src/tcti-tabrmd.c
```diff
@@ -433,7 +433,7 @@ typedef struct {
     GBusType type;
 } bus_name_type_entry_t;
 
-const static bus_name_type_entry_t bus_name_type_map[] = {
+static const bus_name_type_entry_t bus_name_type_map[] = {
```
flihp: This is a good fix. I remember the order of these two qualifiers being unimportant but it looks like C is now obsoleting storage class specifiers anywhere in declarations other than at their beginning.

src/tcti-tabrmd.c
```diff
@@ -641,11 +641,11 @@ Tss2_Tcti_Tabrmd_Init (TSS2_TCTI_CONTEXT *context,
     g_clear_pointer (&conf_copy, g_free);
     g_clear_error (&error);
 
-    return rc;
+    return TSS2_RC_SUCCESS;
```
flihp: Which of the new flags prompted this change?

test/integration/auth-session-start-save.int.c
```diff
@@ -43,7 +43,7 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
 {
     TSS2_RC               rc;
     TPMI_SH_AUTH_SESSION  session_handle = 0;
-    TPMS_CONTEXT          context = { 0, };
+    TPMS_CONTEXT          context = { 0, 0, 0, { 0, 0 } };
```
flihp: Given that this has showed up 4 times now I think it deserves an init macro.

test/integration/common.c
```diff
@@ -41,14 +41,17 @@ create_primary (TSS2_SYS_CONTEXT *sapi_context,
...
+        { 0 } }, { 0, 0 } } };
+    TPM2B_DATA             outside_info    = { 0, 0 };
+    TPML_PCR_SELECTION     creation_pcr    = { 0, { 0, 0, 0 } };
+    TPM2B_PUBLIC           out_public       = { 0, { 0, 0, 0, { 0, 0 }, { 0,
+        { 0 } }, { 0, 0 } } };
+    TPM2B_CREATION_DATA    creation_data   = { 0, { { 0, { 0, 0, 0 } },
+        { 0, 0 }, 0, 0, { 0, 0 }, { 0, 0 }, { 0, 0 } } };
     TPM2B_DIGEST           creation_digest = TPM2B_DIGEST_STATIC_INIT;
-    TPMT_TK_CREATION       creation_ticket = { 0 };
+    TPMT_TK_CREATION       creation_ticket = { 0, 0, { 0, 0 } };
```
flihp: init macros please, pretty please 🙏

src/access-broker.c
dantpm: OK, but there is no mention at the link about whitespace, or at the links at the link.
Just trying to make the code more readable.

src/access-broker.c
dantpm: OK. I can add a static const or macro for this.

src/random.c
dantpm: OK. I found the problem by adding the CCFLAG. I'll add a note to the commit message.

src/tcti-tabrmd.c
dantpm: The compiler complained and warned about "const static". Changing to "static const" stopped the warning.

src/tcti-tabrmd.c
dantpm: It should be "return rc". Thanks.

test/integration/auth-session-start-save.int.c
dantpm: I'll add an init macro, although it really should be in the header file (which is a separate repository).

test/integration/common.c
dantpm: Again, the init macros belong in a separate repository, but I can add them here in abrmd instead.

src/access-broker.c
dantpm: I added TPMS_CAPABILITY_DATA_DEFAULT_INIT to the top of the file.

src/random.c
dantpm: Removed, although it's unclear what the whitespace rule is here. Whitespace between functions makes it more readable.

src/random.c
dantpm: Removed the blank line to make the code more unreadable :-(.

src/random.c
dantpm: Reply is same as above--it is mentioned in the commit message and I expanded the text in the commit message.
FYI, the block comment is *NOT* correct and I corrected the comment.
This problem was discovered when I added -Wextra to CCFLAGS.

src/random.c
dantpm: Done.  Removed line between functions to make it less readable.

dantpm:
Merged with upstream changes.  All issues addressed. Thanks for the review!

src/random.c
flihp: Gotcha, hard to return -1 when you're returning an unsigned number :\

Makefile.am
```diff
@@ -13,7 +13,7 @@ AM_CFLAGS = $(EXTRA_CFLAGS) \
     $(TSS2_SYS_CFLAGS) $(TCTI_DEVICE_CFLAGS) $(TCTI_SOCKET_CFLAGS) \
     $(CODE_COVERAGE_CFLAGS)
 AM_LDFLAGS = $(EXTRA_LDFLAGS) $(CODE_COVERAGE_LIBS)
-UNIT_AM_CFLAGS = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
+UNIT_AM_CFLAGS = $(AM_CFLAGS) $(CMOCKA_CFLAGS) -Wno-unused-parameter
```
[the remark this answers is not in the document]
dantpm: I added it specifically to the files where there are functions with unused parameters.  Usually because of a fixed prototype.
An alternative is to disable it globally.

Per our chat session, I will add a macro to the functions where it appears to force the parameter to be used.  Something like:
#define UNUSED_PARAM(p) (void)(p);

dantpm:
I addressed the latest issue, disabling unused parameter warnings with #define UNUSED_PARAM(p), and all earlier issues.

dantpm:
Updated newest code, integrated after this pull request, to also pass cc -Wextra checks.

This should be ready to integrate.

PR #409, opened 2018-03-30
Fix sessions max

flihp:


dantpm:
This appears to remove this test from the expected XFAIL list:
    test/integration/tcti-sessions-max.int
And rename the test to this (and not expected to fail):
    test/integration/tcti-connections-max.int

PR #411, opened 2018-04-02
Tcti init cleanup

flihp:
The two patches in this PR remove the two legacy init functions: `tss2_tcti_tabrmd_init` and `tss2_tcti_tabrmd_init_full`. Once these are removed then the only supported initialization mechanism is the "standard" one: `Tss2_Tcti_Tabrmd_Init`.

### DIFF
--- a/src/tcti-tabrmd.c
+++ b/src/tcti-tabrmd.c
@@ -514,16 +514,6 @@ tss2_tcti_tabrmd_init_full (TSS2_TCTI_CONTEXT      *context,
     return TSS2_RC_SUCCESS;
 }
 
-TSS2_RC
-tss2_tcti_tabrmd_init (TSS2_TCTI_CONTEXT *context,
-                       size_t            *size)
-{
-    return tss2_tcti_tabrmd_init_full (context,
-                                       size,
-                                       TCTI_TABRMD_DBUS_TYPE_DEFAULT,
-                                       TCTI_TABRMD_DBUS_NAME_DEFAULT);
-}
-
 typedef struct {
     TCTI_TABRMD_DBUS_TYPE type;
     char *name;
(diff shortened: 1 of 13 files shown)


flihp:
Added another commit to update the man page for the relevant function.

src/tcti-tabrmd.c
```diff
@@ -612,46 +513,135 @@ tabrmd_conf_parse (char *conf_str,
...
+    TSS2_TCTI_TABRMD_PROXY (context) =
+        tcti_tabrmd_proxy_new_for_bus_sync (tabrmd_conf.bus_type,
+                                            G_DBUS_PROXY_FLAGS_NONE,
+                                            tabrmd_conf.bus_name,
+                                            TABRMD_DBUS_PATH,
+                                            NULL,
+                                            &error);
+    if (TSS2_TCTI_TABRMD_PROXY (context) == NULL) {
+        g_critical ("failed to allocate dbus proxy object: %s", error->message);
+        rc = TSS2_TCTI_RC_NO_CONNECTION;
```
tstruk: There is something not right here. rc is assigned an error code, which is then ignored and TSS2_RC_SUCCESS is returned.

src/tcti-tabrmd.c
```diff
@@ -612,46 +513,135 @@ tabrmd_conf_parse (char *conf_str,
...
+    rc = tcti_tabrmd_connect (context);
+    if (rc == TSS2_RC_SUCCESS) {
+        g_debug ("initialized tabrmd TCTI context with id: 0x%" PRIx64,
+                 TSS2_TCTI_TABRMD_ID (context));
+    }
+out:
+    g_clear_pointer (&conf_copy, g_free);
+    g_clear_error (&error);
+
+    return TSS2_RC_SUCCESS;
```
tstruk: Should this be:
return rc;

flihp:
never too late, thanks for taking a look.

PR #413, opened 2018-04-03
tcti: Generalize the key / value parsing from the conf string.

flihp:
This commit moves the basics of parsing a string of key / value pairs of
the form "key0=value0,key1=value1,..." from the TCTI module to the util
module. This is the part that parses on "=" and ",". To make this
generic we now provide a callback function that the parsing function
will call passing each parsed key / value pair and a reference to some
caller-provided data.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -340,3 +340,69 @@ g_debug_tpma_cc (TPMA_CC tpma_cc)
     g_debug ("  V:            %s", prop_str (tpma_cc & TPMA_CC_V));
     g_debug ("  Res:          0x%" PRIx8, (tpma_cc & TPMA_CC_RES_MASK) >> TPMA_CC_RES_SHIFT);
 }
+/*
+ * Parse the provided string containing a key / value pair separated by the
+ * '=' character.
+ * NOTE: The 'kv_str' parameter is not 'const' and this function will modify
+ * it as part of the parsing process.
+ */
+gboolean
+parse_key_value (char *key_value_str,
+                 key_value_t *key_value)
+{
+    const char *delim = "=";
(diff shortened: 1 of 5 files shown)


src/util.c
```diff
@@ -340,3 +340,65 @@ g_debug_tpma_cc (TPMA_CC tpma_cc)
...
+ */
+gboolean
+parse_key_value (char *key_value_str,
+                 key_value_t *key_value)
+{
+    const char *delim = "=";
+    char *tok, *state;
+
+    g_debug ("%s: with key_value_str: \"%s\" and key_value_t: 0x%" PRIxPTR, __func__, key_value_str, (uintptr_t)key_value);
+    tok = strtok_r (key_value_str, delim, &state);
```
dantpm: Using strtok_r or strtok is dangerous as it destroys the input string (it replaces the delimiters with the NUL character).
This is OK as long as the caller expects it.  So I would add a note that the input string is modified around line 345.

flihp:
The public API has all `const char*` parameters so no user provided strings are harmed. Internally we make a copy so that we can use `strtok` which as you note is destructive. Adding a note to the comment block above the function isn't a bad idea. I'll fix it up.

src/util.c
flihp: I see your point. The input parameter isn't `const` so the function prototype is clear on this point but lots of people assume otherwise. Adding a comment is easy enough.

dantpm: Thanks. Unfortunately, many C programmers are forget to add const, since it is a "new" feature (~20 years old) or are not always careful.  So the absence of const isn't always that useful.

Issue #412, opened 2018-04-03
Move udev rules to tpm2-tss


martinezjavier:
The udev rules to change the owner and permissions bits for TPM character devices should be in the tpm2-tss project and not in tpm2-abrmd, since users may want to use the device TCTI directly without even installing the tpm2-abrmd.

PR #426, opened 2018-04-13
build: don't install a TPM udev rule

martinezjavier:
This should be done in the tpm2-tss instead, since conceptually the lower
layers of the stack are the ones that access the TPM character devices.

Also, users may want to access the TPM directly without using a resource
manager in user-space.

Fixes: #412

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -145,30 +145,6 @@ started by systemd on boot. If you wish for the daemon to be disabled by
 default some reason you may use this option to the `configure` script to do
 so.
 
-### udev Rules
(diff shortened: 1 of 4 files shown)


AndreasFuchsSIT:
As mentioned in https://github.com/tpm2-software/tpm2-tss/pull/963 the tpm2-tss does not care about users and groups. Only the abrmd actually knows anything about users and groups.

IMHO belongs to here.

martinezjavier:
@AndreasFuchsSIT yes, I answered too. Let's have this discussion there since this change depends on https://github.com/tpm2-software/tpm2-tss/pull/963.

PR #430, opened 2018-04-23
test: Remove bad NULL struct from CMUnitTest array.

flihp:
The CMUnitTest type only has 4 membbers and the NULL struct here has 5
members. This isn't required anyways so the cleanest thing to do is to
just remove it.


flihp:
This fixes build breakage for GCC versions > 5 (the current used by travis) introduced by #407. This will likely be pushed quickly to unblock work.

PR #431, opened 2018-04-23
tcti: Set version field to 2 to correspond with public review TCTI spec.

flihp:
The conf_help string is also updated in the INFO structure to describe
the format of the 'conf' parameter.


flihp:
the build will break here till https://github.com/tpm2-software/tpm2-tss/pull/979 is merged

williamcroberts:
I'm going to build the tools against this to get them ready as well.

Issue #445, opened 2018-05-12
flag check macros don't work


flihp:
The command invoked by the `AX_CHECK_*_FLAG` macros causes clang to report a warning, not an error. This means that the check for a supported flag will always pass. To cause it to fail we must pass `-Werror` to the check via the 3rd parameter.

PR #446, opened 2018-05-12
build: Treat warnings as errors when checking compiler for flag support.

flihp:
The command executed by the AX_CHECK_*_FLAG macro will cause clang to
report a warning when a flag isn't supported. But if we pass this flag
to the build the compilation will fail. So we have to force the warning
to be treated as an error when AX_CHECK_*_FLAG is executed by adding
-Werror to CFLAGS for the check.


flihp:
this resolves #445

Issue #447, opened 2018-05-12
scan-build fails, ci shows build as passing though


flihp:
I didn't catch this in the review but it looks like the scan-build integration with travis-ci never really worked. The results from the first build in master with scan-build turned on fails to compile but the build shows up green in CI regardless. The failing line is here: https://travis-ci.org/tpm2-software/tpm2-abrmd/jobs/374999260#L2001

I'm only noticing this after working on #445

flihp:
Since this prevents the clang build from running properly I'm reverting the offending commit f05f2a850efc88b99e87da31a2316053fe0fe45f. This will be fixed and re-applied once it's fixed and the scan-build findings resolved.

PR #450, opened 2018-05-14
Revert "Add scan-build Static Code Analysis to Travis"

flihp:
This reverts commit f05f2a850efc88b99e87da31a2316053fe0fe45f. The issue
with this commit is that, while it enables scan-build, it overwrites the
contents of CFLAGS and causes the build to fail (all -I directives to
the compiler have been lost and so it can't find the TSS@ headers). Odly
enough, because scan-build didn't find any errors (the build failed so
no errors to find) it still reports success. This isn't what we want.

The right solution here will preserve CFLAGS. A simple change like this
however causes the build to pass complete, scan-build to run, but it
then finds a pile of issues and so the build fails. For the short-term
we're reverting this commit and will resubmit it once the issues
identified by scan-build are resolved.


flihp:
this resolves #447

PR #455, opened 2018-05-15
ci: Use 01.org mirror instead of twobit.us.

### LINKED ISSUE #454: switch mirrors from twobit.us to download.01.org
pretty straight forward

flihp:
Took some time but finally jumped through the hoops required to get
Intel to mirror these dependencies. Should insulate us from upstream
outtages and make us better OSS citizens.

This resolves #454 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,12 +49,12 @@ env:
 install:
   - pip install --user cpp-coveralls
   - mkdir ${DESTDIR} ${DEPS} && pushd ${DEPS}
-  - wget http://ftpmirror.gnu.org/autoconf-archive/autoconf-archive-2017.09.28.tar.xz
+  - wget https://download.01.org/tpm2/autoconf-archive-2017.09.28.tar.xz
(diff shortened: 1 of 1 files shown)


williamcroberts:
> Took some time but finally jumped through the hoops required to get Intel to
> mirror these dependencies. Should insulate us from upstream outtages and
> make us better OSS citizens.

All my deps are on the docker image, which simplifies things at the expense of having to pull
more data. It looks like dockerhub is built for that type of paradigm (pull for CI).

Not sure if I should throw my docker image on this mirror...thoughts?

flihp:
Dunno. I think you're right as far as dockerhub goes but I don't know too much about it. Getting 01.org to host shouldn't be an issue though (depending on size I guess). Out of curiosity: how big is the docker image?

williamcroberts:
> Dunno. I think you're right as far as dockerhub goes but I don't know too much
> about it. Getting 01.org to host shouldn't be an issue though (depending on size I
> guess). Out of curiosity: how big is the docker image?

396 MB

flihp:
can't hurt to ask I guess?

PR #487, opened 2018-07-03
Update man pages to reflect changes in tcti-mssim configuration strings.

safayetahmedatge:


flihp:
Good catch, thanks @safayetahmedatge.

Issue #490, opened 2018-07-10
race condition in session mgmt logic


flihp:
I haven't been able to trigger this locally but travis seems to have the magic. Getting the contents of the test logs from travis is less than simple though so I have a pile of test code and cleanup in the 'ci-logs' branch where I'm collecting data. The relevant log can be found here: https://api.travis-ci.org/v3/job/402346519/log.txt and the failing test is: session-load-from-closed-connections-lru.

Interesting enough the race can happen for a number of the integration tests that exit immediately after using a session so YMMV.

flihp:
The error text is: ** (tpm2-abrmd:32591): ERROR **: Connection closed with session in unexpected state: loaded

flihp:
this was resolved in #492

PR #492, opened 2018-07-11
ResourceManager: Fix race between session save and connection termination.

flihp:
0115d572bc4f6026a28a2ed5d7340ca77c6dc4d2 fixed the worst part of this
but a very narrow window for a race persisted. For it to trigger the
client must terminate their connection to the daemon triggering a call
to the resource_manager_on_connection_removed function by an external
thread before the main processing thread saves all of the sessions
associated with the previous command.

Fixing this requires that the main processing loop hold the lock on the
SessionList until all sessions have been saved.


flihp:
this resolves #490

PR #493, opened 2018-07-11
test: Work around scan-build false positive with simple assert.

flihp:
scan-build doesn't understand the cmocka asserts. Without this patch
it logs a false positive:

test/resource-manager_unit.c:383:17: warning: 1st function call argument is an uninitialized value
        entry = handle_map_entry_new (phandles [i], vhandles [i]);
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.

As a work-around we can use a conditional and a standard 'assert' which
scan-build understands just fine. There's probably a more elegant way
to inform scan-build about the cmocka asserts but I haven't found anything
concrete yet.

### DIFF
--- a/test/resource-manager_unit.c
+++ b/test/resource-manager_unit.c
@@ -24,6 +24,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
  * THE POSSIBILITY OF SUCH DAMAGE.
  */
+#include <assert.h>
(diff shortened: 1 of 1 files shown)


flihp:
Worth noting is that this failure went undetected by our CI. scan-build was set up properly and reporting the false-positive the code integrating scan-build was however ignoring the results. PR to fix this is forthcoming.

PR #494, opened 2018-07-11
Ci cleanup

flihp:
While debugging #490 it became plane that the current CI script needed some cleanup. Part of this was motivated by the need to just getting meaningful output from the test harness logs when a test fails. While doing this I realized that the CI configuration was also ignoring the results from scan-build which was suppressing a finding from the tool (granted it was a false positive). These two patches fix the issue of the logs remaining invisible when a test fails as well as treating findings from scan-build as failures. 

I've also rolled in some efficiency gains by removing unnecessary build steps that we were using as a crutch. Removing these steps reduces our build time (as reported by travis-ci) from ~12 minutes to ~9 which is pretty significant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,18 +86,27 @@ script :
     fi
   - |
     CONFIGURE_OPTIONS="--disable-dlclose --enable-unit --enable-integration"
-    if [[ "$CC" == clang* ]]; then
+    case "$CC" in
+    clang*)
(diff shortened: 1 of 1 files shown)


flihp:
the CI failure here is expected since this PR fixes a bug that effectively ignored findings from scan-build. Once #493 is merged this should turn green.

Issue #475, opened 2018-06-18
TEMP_FAILURE_RETRY is specific to the GNU C Library


safayetahmedatge:
https://github.com/tpm2-software/tpm2-abrmd/blob/c90cd599bbde05514a6396ad25c47aec74bbd70a/src/tcti-tabrmd.c#L168

This is problematic on platforms that use a non-GNU C library like musl libc in Alpine containers. The musl libc headers don't implement TEMP_FAILURE_RETRY.

Other projects that see similar issues address it with preprocessor directives similar to the following:
```
#ifndef TEMP_FAILURE_RETRY
#define TEMP_FAILURE_RETRY(expression) \
...
```
We could do the same in 'src/util.h'. My preference would be to just copy the implementation from an existing project but I am unsure of the licensing restrictions. Can tpm2-abrmd include Apache licensed code like the following?

https://android.googlesource.com/platform/system/core/+/master/base/include/android-base/macros.h#26

flihp:
We have an implementation of this paradigm in tpm2-tss for handling retries on Windows and Linux. Should be easy enough to just use that code.

safayetahmedatge:
Thanks @flihp. Will look into it.

PR #497, opened 2018-07-17
Remove dependence on GNU libc to improve portability

safayetahmedatge:


safayetahmedatge:
Is there a requirement that the code should be consistent across machines with different endianness? I'm not entirely sure that the changes to "random_get_bytes" will meet such a requirement. I'm open to suggestions on a different way to do it.

flihp:
Thanks for the PR @safayetahmedatge. Before I do the review a minor procedural thing: please do look at the CONTRIBUTING.md file for patch submission guidelines. Specifically neither of these commits meet the DCO requirement.

Additionally I'd recommend you break this PR up into two, one for each commit. I recommend this because the first commit is relatively uncontroversial since it was discussed in #475 and should be something we can get committed with a quick review. The second commit is something we haven't discussed in an issue thread yet and will take a bit of research on my part to understand what you're changing, why / how etc.

Issue #237, opened 2017-11-08
support for minimal libc


flihp:
Distros like Angstrom use musl libc instead of the GNU libc and there are a few compatibility issues. Angstrom packages glib so I'm assuming whatever non-standard features we're using from GNU libc can be worked around.

flihp:
this was resolved thanks to @safayetahmedatge in #503

PR #502, opened 2018-07-18
Remove TEMP_FAILURE_RETRY call for portability

safayetahmedatge:
TEMP_FAILURE_RETRY is a non-protable extension to GNU libc.

This PR replaces it with implementation similar to TSS2_RETRY_EXP in
tpm2-tss, but consistent with behaviour described in the GNU C library manual:
  https://www.gnu.org/software/libc/manual/html_node/Interrupted-Primitives.html

Adressing issue https://github.com/tpm2-software/tpm2-abrmd/issues/475

### DIFF
--- a/src/tcti-tabrmd.c
+++ b/src/tcti-tabrmd.c
@@ -165,7 +165,7 @@ tcti_tabrmd_poll (int        fd,
     int ret;
     int errno_tmp;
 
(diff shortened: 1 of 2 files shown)


flihp:
Great patch @safayetahmedatge. The commit message to comment ratio is much better this time around.

flihp:
also relevant to #237

Issue #514, opened 2018-08-27
Upgrading libblkid1 from 2.30.2 to 2.31.1 causes cryptfs-tpm2 to error with "Too large count value passed to g_input_stream_read"


gblewis1:
See https://github.com/jiazhang0/cryptfs-tpm2/issues/3 for details.  I've identified a specific commit in util-linux (https://github.com/karelzak/util-linux/commit/d2b0c6587d729a8e838821cb70d4de18c6969868) that appears to interfere with communication between cryptfs-tpm2 and tpm2-abrmd.  I'm unsure where this issue belongs.

flihp:
I don't know where this issue belongs either unfortunately. The warning you're seeing is coming from here: https://github.com/tpm2-software/tpm2-abrmd/blob/master/src/util.c#L176. The source of the error is the call to `g_input_stream_read` which is how we get data from clients. So the error message
> Too large count value passed to g_input_stream_read
is coming from libglib. After this error is detected the client connection that caused the error is closed. This is why your client is being disconnected.

Can you please reproduce this with the glib logging dialed up. You can do this by setting `G_MESSAGES_DEBUG=all` in the environment.

flihp:
I'm just going with my gut here but this looks like an integer underflow here: https://github.com/tpm2-software/tpm2-abrmd/blob/master/src/util.c#L170

gblewis1:
With G_MESSAGES_DEBUG=all, on the cryptfs-tpm2 side I am getting:

```
Tue Aug 28 18:16:32 UTC 2018: [INFO] Use tabrmd as the default tcti interface
** (process:20547): DEBUG: initialized tabrmd TCTI context with id: 0x27698b7171430a6
** (process:20547): DEBUG: tss2_tcti_tabrmd_transmit
** (process:20547): DEBUG:     80 01 00 00 00 16 00 00 01 7a 00 00 00 05 00 00
** (process:20547): DEBUG:     00 00 00 00 00 01
** (process:20547): DEBUG: blocking write on iostream: 0x7f2028007ad0
** (process:20547): DEBUG: writing 22 bytes starting at 0x1372628 to socket 0x7f2033855ec0
** (process:20547): DEBUG: wrote 22 bytes to ostream 0x7f2028007ad0
** (process:20547): DEBUG: returning 22
** (process:20547): DEBUG: tss2_tcti_tabrmd_receive
** (process:20547): DEBUG: poll has 1 fds ready
[...]

PR #519, opened 2018-08-29
util: Fix sign of format specifier in debug message.

flihp:
The value being printed is unsigned but the conversion specifier zd is
used. zd is for signed sizes (aka ssize_t).


flihp:
This is related to #514 in that it is the cause of some weird debug output that makes understanding the issue more difficult.

flihp:
needs to be cherry-picked into 1.x too

PR #543, opened 2018-09-26
Handle ContextSave and ContextLoad operations for sessions.

flihp:
This commit contains structural changes that are required for handling
the context gap response code. The previous approach was as passive as
possible when it came to handling sessions. The ResourceManager passed
all ContextSave and ContextLoad commands through to the TPM and
extracting the data it needed from them as they passed. This data was
sufficient for us to load and save contexts when clients referenced them
in the handle / auth area of a command.

This design works so long as we don't need to load / save a session
context in between a ContextSave & ContextLoad from the Connection that
owns it. We previously worked around this by refusing to do this. This
was fine since the TPM throws errors if it receives a command that
references sessions that aren't loaded and we were just preserving this
behavior.

This design falls down if we ever encounter a situation in which we
*must* break the rules described above. Handling the context gap is
exactly this situation. This commit changes this design as follows:

SessionEntry: This is where we store the context object when it's saved.
We now maintain two copies of the context object. The first is the copy
that the ResourceManager uses, the second is the one that we expose to
the client. These TPMS_CONTEXT structures are stored in their marshalled
form. We also add a few utility functions to make handling these blobs
easier.

tpm2-header: This commit adds a convenience function to this module for
setting the header fields in a memory blob. This is very useful when
we're constructing response buffers to send back to the caller on behalf
of the TPM.

Tpm2Command: We've added two new functions to create TPM2_ContextLoad
and TPM2_ContextSave commands. Now that we're storing the TPMS_CONTEXT
structures as marshalled blobs we use a different interface in the
AccessBroker to send Tpm2Command objects that we generate manually.

Tpm2Response: We've added two new functions to create Tpm2Response
objects suitable for responding to clients executing the
TPM2_ContextSave and TPM2_ContextLoad commands. This includes creating
the response buffer body from the TPM_CONTEXT structure / handle where
appropriate.

SessionList: This object needded a way to lookup SessionEntry objects
based on the provided context blob. This uses comparison mechanisms
added to the SessionEntry object.

ResourceManager: The changes above allow us to handle a.k.a.
"virtualize" the ContextSave and ContextLoad operations for sessions.
This required a lot of refactoring and some new code to intercept these
commands and generate an appropriate response. All code associated with
transferring session ownership has been relocated to these functions.

### DIFF
--- a/src/access-broker.c
+++ b/src/access-broker.c
@@ -496,7 +496,7 @@ access_broker_send_command (AccessBroker  *broker,
     g_debug ("access_broker_send_command: AccessBroker: 0x%" PRIxPTR
              " Tpm2Response: 0x%" PRIxPTR " RC: 0x%" PRIx32, (uintptr_t)broker,
              (uintptr_t)response, tpm2_response_get_code (response));
-    g_object_unref (connection);
+    g_clear_object (&connection);
     return response;
 
 unlock_out:
--- a/src/resource-manager.c
+++ b/src/resource-manager.c
@@ -26,8 +26,11 @@
  */
 #include <errno.h>
 #include <inttypes.h>
+#include <string.h>
 
 #include <glib.h>
+
+#include <tss2/tss2_mu.h>
 
 #include "connection.h"
 #include "connection-manager.h"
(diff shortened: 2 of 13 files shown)


flihp:
Also worth noting is that only the existing tests were updated. Adding this code will bring down our test stats a bit. Will be adding additional test coverage during 2.1 work.

Issue #541, opened 2018-09-25
CRITICAL **: failed to allocate dbus proxy object: Error calling StartServiceByName for com.intel.tss2.Tabrmd: Timeout was reached


jturnsek:
#include <errno.h>
#include <glib.h>
#include <inttypes.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

#include "tss2-tcti-tabrmd.h"

#define CONNECTION_COUNT 2

TSS2_RC
tcti_tabrmd_init (TSS2_TCTI_CONTEXT **tcti_context,
                  const char         *conf,
                  int                 retries)
{
    TSS2_RC rc;
    size_t size, i;

    rc = Tss2_Tcti_Tabrmd_Init (NULL, &size, NULL);
    if (rc != TSS2_RC_SUCCESS) {
        fprintf (stderr, "Failed to get allocation size for tabrmd TCTI "
                 " context: 0x%" PRIx32 "\n", rc);
        return rc;
    }
    *tcti_context = calloc (1, size);
    if (*tcti_context == NULL) {
        fprintf (stderr, "Allocation for TCTI context failed: %s\n",
                 strerror (errno));
        return rc;
    }
    for (i = 0; i < (size_t) retries; ++i) {
        rc = Tss2_Tcti_Tabrmd_Init (*tcti_context, &size, conf);
        if (rc == TSS2_RC_SUCCESS) {
            break;
        } else {
            g_info ("Failed to initialize tabrmd TCTI context: 0x%" PRIx32
                    " on try %zd", rc, i);
            sleep (1);
        }
    }
    if (rc != TSS2_RC_SUCCESS) {
        fprintf (stderr, "Failed to initialize tabrmd TCTI context: "
                 "0x%" PRIx32 "\n", rc);
        free (*tcti_context);
    }
    return rc;
}

int
main (int   argc,
      char *argv[])
{
    pid_t childPID;
    TSS2_RC rc;
    TSS2_TCTI_CONTEXT *tcti_context[CONNECTION_COUNT] = { 0 };
    size_t i;
    
    rc = tcti_tabrmd_init (&tcti_context [0],
                           NULL,
                           3);
    if (tcti_context [0] == NULL || rc != TSS2_RC_SUCCESS) {
        g_error ("failed to connect to TCTI: 0x%" PRIx32, rc);
    }

    childPID = fork();

    rc = tcti_tabrmd_init (&tcti_context [1],
[...]

PR #544, opened 2018-10-11
Fork tests

flihp:
This PR is two tests that cover scenarios where the test creates TCTI connection initialization before and after the 'fork' systemcall. This is part of work toward a resulution for #541


flihp:
Odd that the error string we get is 'Timeout exceeded' but the stalled travis builds seem to indicate that the timeout value is different between two instances of the traivs build image: gcc test fails in under 5 minutes, clang build hangs indefinitely, the opposite is true when built by travis from my github space o_O

flihp:
All signs currently point to this being a glib issue that we can't fix without removing the glib dependency on the client side. No use in merging tests for stuff we can't fix.

PR #546, opened 2018-10-12
Simplify autotools machinery

joshuagl:
We don't need to explicitly check for the existence of gdbus-codegen, nor store its path. It should be safe to assume that the binary is available on `PATH` for any system with the gio pkg-config files available.


flihp:
First commit here is good, second one undoes the goodness from the first one and a bunch of other checks. The second commit looks like it was done in anger really :smile: Closing this PR, will have another open with the first commit and a bit of cleanup.

Issue #551, opened 2018-10-31
account for tpm2-tss build using OpenSSL by default


flihp:
The travis-ci buidl is currently failing on master. This is caused by an upstream change. Quick fix is to build tpm2-tss using gcrypt. Long term fix is to build tpm2-tss using openssl on travis-ci.

flihp:
Using the libssl-dev from trusty (ubuntu version travis-ci uses) doesn't work. Looks like it's v1.0.1 and we need at least v1.0.2. This is likely why the tpm2-tss travis-ci config downloads and builds v1.0.2 the hard way. Try looking in trusty-backports first for v1.0.2. If it's not there fall back to the method used by tpm2-tss (last resort).

flihp:
This was resolved when the travis-ci instance was upgraded to `xenial` in #561.

PR #552, opened 2018-10-31
travis-ci: Build tpm2-tss using gcrypt.

flihp:
Upstream the tpm2-tss is now defaulting to use openssl as its crypto
library. Supporting this will require a few changes to the ci build.
Best short-term fix is to keep using grcypt until ci is update.


flihp:
Merging this quickly to fix the build. See #551 for additional data / strategy to get this fixed "the right way".

PR #556, opened 2018-11-23
build: improve detection of stack protector support

casantos:
Some toolchains provide compilers that recognize -fstack-protector-all
but do not actually support SSP in their system libraries, leading to
link-time errors. Prevent this by means of a toolchain-wide test that
attempts to link a program compiled with -fstack-protector-all.

Fixes:
  http://autobuild.buildroot.net/results/bd9005eeb24678aa530179a80bbc99b2176f8559
  http://autobuild.buildroot.net/results/feff61dcb481a94f5f030117830984c5e09727ea

Suggested-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>
Signed-off-by: Carlos Santos <casantos@datacom.com.br>


flihp:
Thanks for rebasing this @casantos. This macro looks like a minor variation on the `AX_ADD_COMPILER_FLAG` macro. The variation being that the flag is checked using `AX_CHECK_LINK_FLAG` instead of `AX_CHECK_COMPILE_FLAG` before it's added to `EXTRA_CFLAGS`. If this is correct would it be possible to achieve the same effect by adding an additional call to `AX_ADD_LINK_FLAG`? I understand that this would add the flag to `EXTRA_LDFLAGS` but that doesn't seem to cause any problems. What do you think?

flihp:
Did some thinking on this and I've convinced myself that my proposed solution won't work. If we use the `AX_CHECK_LINK_FLAG` macro the check  will fail and that flag won't be added to `LDFLAGS` but the call to `AX_CHECK_COMPILER_FLAG` will still pass and `stack-protect-all` will still be added to `CFLAGS`. This will leave us right back where we started.

Thanks for playing along while I got my head around this.

PR #557, opened 2018-11-23
Fix Travis CI build

diabonas:
Since https://github.com/tpm2-software/tpm2-tss/pull/1199 was merged upstream, the Gnulib include directory [needs to be specified during bootstrap of tpm2-tss](https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md#bootstrapping-the-build).

Also remove the doxygen package, which is not necessary any more since https://github.com/tpm2-software/tpm2-tss/pull/1209 was merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
       - dbus
       - dbus-x11
       - doxygen
+      - gnulib
       - cmake
       - lcov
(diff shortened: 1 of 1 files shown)


casantos:
#556 is waiting for this one.

PR #563, opened 2018-12-07
Cleanup

flihp:
A pile of odds and ends that I ran across while working on new features.

### DIFF
--- a/src/resource-manager.c
+++ b/src/resource-manager.c
@@ -373,6 +373,15 @@ resource_manager_save_session_context (gpointer data_entry,
     }
     resp = access_broker_send_command (resmgr->access_broker, cmd, &rc);
     if (rc != TSS2_RC_SUCCESS) {
+        g_critical ("%s: TCTI failed while saving session context from "
+                   "SessionEntry 0x%" PRIxPTR ", got RC 0x%" PRIx32,
+                   __func__, (uintptr_t)entry, rc);
+        goto err_out;
+    }
+    if (tpm2_response_get_code (resp) != TSS2_RC_SUCCESS) {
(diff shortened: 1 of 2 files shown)


src/resource-manager.c
```diff
@@ -373,6 +373,15 @@ resource_manager_save_session_context (gpointer data_entry,
     }
     resp = access_broker_send_command (resmgr->access_broker, cmd, &rc);
     if (rc != TSS2_RC_SUCCESS) {
+        g_critical ("%s: TCTI failed while saving session context from "
+                   "SessionEntry 0x%" PRIxPTR ", got RC 0x%" PRIx32,
+                   __func__, (uintptr_t)entry, rc);
+        goto err_out;
+    }
+    if (tpm2_response_get_code (resp) != TSS2_RC_SUCCESS) {
```
joshuagl: Any reason why one is a g_critical and the other a g_warning? g_critical is usually used for programmer errors, but I can't tell from the diff context whether this is one?

src/resource-manager.c
```diff
@@ -665,14 +665,10 @@ resource_manager_flush_context (ResourceManager *resmgr,
...
-        g_info ("f");
+        g_debug ("%s: handle 0x%08" PRIx32 "is a session, removing from "
+                 "SessionList", __func__, handle);
         session_list_remove_handle (resmgr->session_list, handle);
-        /*
-        response = access_broker_send_command (resmgr->access_broker,
-                                               command,
-                                               &rc);
-        */
+        session_list_prettyprint (resmgr->session_list);
```
joshuagl: I had to dig a few levels deep to find out the `pretty_print*` functions were using `g_debug`, and therefore require the _G\_DEBUG_ environment variable to be set. LGTM.

PR #565, opened 2018-12-09
handle TPM2_RC_CONTEXT_GAP

flihp:
This PR contains
- a test case to trigger and test for the proper handling of the context gap response code
- a bit of refactoring to cleanup the session context management functions in the ResourceManager object
- the logic necessary to handle the context gap response code on behalf of clients

### DIFF
--- a/src/resource-manager-session.c
+++ b/src/resource-manager-session.c
@@ -135,3 +135,49 @@ out:
     g_clear_object (&cmd);
     return resp;
 }
+/*
+ * This function will send the required commands in order to bring a
+ * session back into the gap interval. This means loading and saving
+ * the associated context. If the SessionEntry is not saved then this
+ * function is a no-op.
+ */
+gboolean
+regap_session (ResourceManager *resmgr,
+               SessionEntry *entry)
+{
+    SessionEntryStateEnum state;
+    Tpm2Response *resp = NULL;
+    gboolean ret = TRUE;
+
+    g_assert_nonnull (resmgr);
+    g_assert_nonnull (entry);
+
+    state = session_entry_get_state (entry);
+    g_debug ("%s: swapping SessionEntry 0x%" PRIxPTR " in state \"%s\"",
+             __func__, (uintptr_t)entry, session_entry_state_to_str (state));
+    if (state == SESSION_ENTRY_SAVED_CLIENT ||
+        state == SESSION_ENTRY_SAVED_CLIENT_CLOSED ||
+        state == SESSION_ENTRY_SAVED_RM)
+    {
+        resp = load_session (resmgr, entry);
+        if (tpm2_response_get_code (resp) != TSS2_RC_SUCCESS) {
+            g_critical ("%s: Failed to save SessionEntry 0x%" PRIxPTR "removing "
+                        "from list", __func__, (uintptr_t)entry);
+            flush_session (resmgr, entry);
+            ret = FALSE;
+            goto out;
+        }
+        g_clear_object (&resp);
+        resp = save_session (resmgr, entry);
+        if (tpm2_response_get_code (resp) != TSS2_RC_SUCCESS) {
+            g_critical ("%s: Failed to load SessionEntry 0x%" PRIxPTR ". Got RC "
+                        "0x%" PRIx32 ", removing from list",
+                        __func__, (uintptr_t)entry, tpm2_response_get_code (resp));
+            flush_session (resmgr, entry);
+            ret = FALSE;
+        }
+    }
+out:
+    g_clear_object (&resp);
+    return ret;
+}
--- a/src/resource-manager-session.c
+++ b/src/resource-manager-session.c
@@ -1,0 +1,137 @@
+/* SPDX-License-Identifier: BSD-2 */
+#include <glib.h>
+#include <inttypes.h>
+
+#include "access-broker.h"
+#include "resource-manager.h"
+#include "resource-manager-session.h"
+#include "session-list.h"
+#include "session-entry.h"
+#include "tabrmd.h"
+#include "tpm2-command.h"
+#include "tpm2-header.h"
+#include "tpm2-response.h"
+#include "util.h"
+
+/*
+ * Load the TPMS_CONTEXT associated with the provided SessionEntry*.
+ * This function will always return a Tpm2Response object (never NULL)
+ * though the response object may indicate a failure in the response code.
+ */
+Tpm2Response*
+load_session (ResourceManager *resmgr,
+              SessionEntry *entry)
+{
+    Tpm2Command *cmd = NULL;
+    Tpm2Response *resp = NULL;
+    size_buf_t *size_buf = NULL;
+    TSS2_RC rc = TSS2_RC_SUCCESS;
+
+    g_debug ("%s: SessionEntry 0x%" PRIxPTR, __func__, (uintptr_t)entry);
+    size_buf = session_entry_get_context (entry);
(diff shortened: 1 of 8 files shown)


src/resource-manager-session.c
```diff
@@ -135,3 +135,46 @@ save_session (ResourceManager *resmgr,
...
+        state == SESSION_ENTRY_SAVED_RM)
+    {
+        resp = load_session (resmgr, entry);
+        if (tpm2_response_get_code (resp) != TSS2_RC_SUCCESS) {
+            g_critical ("%s: Failed to save SessionEntry 0x%" PRIxPTR "removing "
+                        "from list", __func__, (uintptr_t)entry);
+            flush_session (resmgr, entry);
+            ret = FALSE;
+        }
+        resp = save_session (resmgr, entry);
```
joshuagl: Are we not potentially trying to call `save_session()` on a session for which `load_session()` failed? Are there any negative consequences of that?

src/resource-manager-session.c
```diff
@@ -135,3 +135,46 @@ save_session (ResourceManager *resmgr,
...
+        if (tpm2_response_get_code (resp) != TSS2_RC_SUCCESS) {
+            g_critical ("%s: Failed to load SessionEntry 0x%" PRIxPTR ". Got RC "
+                        "0x%" PRIx32 ", removing from list",
+                        __func__, (uintptr_t)entry, tpm2_response_get_code (resp));
+            flush_session (resmgr, entry);
+            ret = FALSE;
+        }
+    }
+    g_clear_object (&resp);
```
joshuagl: Are we potentially leaking an object here? We've assigned to resp twice, once with `load_session()` and once with `save_session()`.

src/resource-manager.c
```diff
@@ -138,6 +138,65 @@ resource_manager_load_transient (ResourceManager  *resmgr,
...
+        g_debug ("%s: handling TPM2_RC_CONTEXT_GAP", __func__);
+        session_list_foreach (resmgr->session_list,
+                              regap_session_callback,
+                              &data);
+        ret = data.ret;
+        break;
+    default:
+        g_debug ("%s: Unable to recover gracefully from RC 0x%" PRIx32,
+                 __func__, rc);
+        ret = TRUE;
```
joshuagl: If we can't gracefully recover, shouldn't we return `FALSE` here so that the calling site can propagate the error? Or are we returning `TRUE` and expecting the second `[load|save]_session()` to keep things sane? Sorry, I haven't had chance to dig into this deeply yet.

joshuagl:
The commit message for the 2nd patch "ResourceManager: Handle context gap event when saving sessions." includes the statement:

> The majority of this commit however is refactoring the ResourceManager
> code that loads and saves sessions. These have been separated out into a
> new code module 'resource-manager-session' and cleaned up.

However that work appears to have been factored out into patch 3?

test/integration/session-gap.int.c
```diff
@@ -0,0 +1,105 @@
...
+    UINT32 gap_max = 0;
+    UINT64 initial_sequence = 0, last_sequence = 0;
+
+    rc = get_context_gap_max (sapi_context, &gap_max);
+    if (rc != TSS2_RC_SUCCESS) {
+        g_error ("%s: get_context_gap_max failed with RC: 0x%" PRIx32,
+                 __func__, rc);
+    }
+    g_info ("%s: GAP_MAX is 0x%" PRIx32, __func__, gap_max);
+    if (gap_max == 0xffffffff) {
```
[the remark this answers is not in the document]
flihp: This should have been removed as part of refactoring the GetCapability part of this work. This PR should have only the code to handle the RC. Interception of GetCapability will be next and I'll move this code over to that branch.

src/resource-manager-session.c
flihp: Good catch. This function was missing a `goto` in that first error handling block.

src/resource-manager-session.c
flihp: good catch

src/resource-manager.c
flihp: By returning `TRUE` here we cause the RC to be sent back to the caller. Just because we don't know how to handle it doesn't mean they won't.

joshuagl: Right, that makes sense. Thanks for clarifying!

flihp: I spent some time looking back over this because my answer felt wrong. And it was. What will happen here if we fail to load the session:
- the session will be flushed
- the callers command that caused the RM to try to load the session will be sent to the TPM
- it will fail since it references a session that we couldn't load
- caller will get an RC from the TPM indicating that one of the sessions it references

If we want to handle everything in the RM we could construct a response object but sending the callers command through to the TPM and letting it create the RC seemed like the right thing to do at the time. Since we have all of the nice things provided by the type marshalling library (libtss2-mu) constructing response objects isn't nearly as much work as it used to be but even then packing out the right RC bits is a bit of work. I'm not against changing the approach here but I'd say it would make more sense in a future PR that makes the behavior consistent (we do similar things elsewhere).

joshuagl: Thanks for looking back and this writing up. I'm OK with doing the work in a future PR, but could we file an enhancement request so that we don't lose this context?

src/resource-manager-session.c
```diff
@@ -0,0 +1,137 @@
...
+                    __func__);
+        resp = tpm2_response_new_rc (NULL, TSS2_RESMGR_RC_GENERAL_FAILURE);
+        goto out;
+    }
+    resp = access_broker_send_command (resmgr->access_broker, cmd, &rc);
+    if (rc != TSS2_RC_SUCCESS) {
+        g_critical ("%s: TCTI failed while loading session context from "
+                   "SessionEntry 0x%" PRIxPTR ", got RC 0x%" PRIx32,
+                   __func__, (uintptr_t)entry, rc);
+        resp = tpm2_response_new_rc (NULL, rc);
```
joshuagl: We reassign resp here, are we leaking the one returned from the `access_broker_send_command()` call above?

src/resource-manager-session.c
```diff
@@ -0,0 +1,137 @@
...
+                    __func__);
+        resp = tpm2_response_new_rc (NULL, TSS2_RESMGR_RC_GENERAL_FAILURE);
+        goto out;
+    }
+    resp = access_broker_send_command (resmgr->access_broker, cmd, &rc);
+    if (rc != TSS2_RC_SUCCESS) {
+        g_critical ("%s: TCTI failed while saving session context from "
+                   "SessionEntry 0x%" PRIxPTR ", got RC 0x%" PRIx32,
+                   __func__, (uintptr_t)entry, rc);
+        resp = tpm2_response_new_rc (NULL, rc);
```
joshuagl: Another leak of a Tpm2Response?

src/resource-manager-session.c
flihp: You'll need to take a look at the `access_broker_send_command` function to understand this. The `Tpm2Response` object returned from the function represents the response from the TPM. The `rc` parameter (an 'out' value / passed by reference) represents the `TPM2_RC` from the TCTI layer. If the TCTI returns an error then the return value will be NULL. Thus the 'resp' value here will not be overwritten / leaked because it is NULL.

src/resource-manager-session.c
flihp: same answer as above

flihp:
Appreciate the review @joshuagl. Think we're at the point of diminishing returns though and I'd like to get work on the GetCapability part of this done before the holiday + a release.

PR #568, opened 2018-12-23
misc: Use SPDX license header.

flihp:
This commit required a bit of awk magic.


flihp:
this resolves #474

PR #572, opened 2019-01-02
Mask object addresses in debug output.

flihp:
Dumping the address of a GObject in debug output is useful as a
mechanism for tracking objects. This effectively uses the address of the
object as an ID value. This is however undesirable as there is a
possibility that a system may be deployed with debug output turned on
mistakenly. This would case the log file to contain info that may be
used to weaken ASLR.

To combat this we add a new function to the util module that we use to
mask object addresses. This is done by combining the address with a
random value using the bitwise exclusive or (aka XOR) operation. The
random value / seed is initialized by adding another new function to
initialize the 'util' module. This function must be called before any
addresses are masked.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -16,6 +16,7 @@
 
 #include <tss2/tss2_tpm2_types.h>
 
+#include "random.h"
 #include "util.h"
 #include "tpm2-header.h"
 
@@ -76,11 +77,8 @@ write_all (GOutputStream *ostream,
     GError *error = NULL;
 
     do {
-        g_debug ("writing %zu bytes starting at 0x%" PRIxPTR " to socket 0x%"
-                 PRIxPTR,
-                 size - written_total,
-                 (uintptr_t)(buf + written_total),
-                 (uintptr_t)socket);
+        g_debug ("%s: writing %zu bytes to ostream", __func__,
+                 size - written_total);
         written = g_output_stream_write (ostream,
                                          (const gchar*)&buf [written_total],
                                          size - written_total,
@@ -89,15 +87,13 @@ write_all (GOutputStream *ostream,
         switch (written) {
         case -1:
             g_assert (error != NULL);
-            g_warning ("failed to write to ostream 0x%" PRIxPTR ": %s",
-                       (uintptr_t)ostream, error->message);
+            g_warning ("%s: failed to write to ostream: %s", __func__, error->message);
             g_error_free (error);
             return written;
         case  0:
             return (ssize_t)written_total;
         default:
-            g_debug ("wrote %zd bytes to ostream 0x%" PRIxPTR,
-                     written, (uintptr_t)ostream);
+            g_debug ("%s: wrote %zd bytes to ostream", __func__, written);
         }
         written_total += (size_t)written;
     } while (written_total < size);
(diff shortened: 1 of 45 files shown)


flihp:
rebased on master after merging #571

src/util.c
```diff
@@ -385,3 +384,35 @@ parse_key_value_string (char *kv_str,
...
+    g_clear_object (&rand);
+}
+
+void*
+objid (const void *ptr)
+{
+    g_debug ("%s: ptr_mask: %p", __func__, ptr_mask);
+    g_assert (ptr_mask != NULL);
+
+    return (void*)((uintptr_t)ptr_mask ^ (uintptr_t)ptr);
```
williamcroberts: I guess in this case, the XOR Cipher is fine since key length == message length and the plain text is fully controlled. But I am no crypto expert.

williamcroberts: If you print a NULL pointer you print the key....

flihp: Since this never happens in the code I didn't consider it an issue. Easy enough to have the `objid` function balk if passed a `NULL` pointer. Will keep me from screwing things up in the future.

src/util.c
```diff
@@ -385,3 +384,23 @@ parse_key_value_string (char *kv_str,
...
+static void* ptr_mask = NULL;
+
+void
+util_init (void)
+{
+    g_assert (ptr_mask == NULL);
+    Random *rand = random_new ();
+    g_assert (random_seed_from_file (rand, ENTROPY_SRC) != -1);
+    g_assert (random_get_bytes (rand, (uint8_t*)&ptr_mask, sizeof (ptr_mask)) == sizeof (ptr_mask));
+    g_clear_object (&rand);
```
williamcroberts: Wouldn't you want to try again got a random number while ptr_mask is NULL?

src/util.c
flihp: w/r/t use of XOR in your first comment: I looked into your initial example of using the kernel and it's `hash` function to mask the address. The built-in `hash` functions from the kernel aren't "cryptographic hash functions" so they're only intended for use in hash tables and the like. Relying on the "one-way" property of these hashes is a "very bad thing".

XOR seemed like a better solution for the reasons you list above. Alternatively using a cryptographic hash is possible but it will significantly increase the size of the output (from sizeof (void*) to 128 bytes in the case of SHA2) and it's more expensive.

williamcroberts: I think you can just return NULL if passed NULL since NULL is never really a usable info leak.

> XOR seemed like a better solution for the reasons you list above. Alternatively using a cryptographic hash is possible but it will significantly increase the size of the output (from sizeof (void*) to 128 bytes in the case of SHA2) and it's more expensive.

It's only expensive on the debug case. I think the xor cipher is fine here, again I am not nor claim to be a crypto expert.

src/util.c
williamcroberts: Yeah so why have it fail late when you can prevent it upfront?

flihp: I'd rather the code fail after getting 4 bytes of `0`s out of `/dev/urandom` since this is an exceptional circumstance. I don't object to retrying some fixed number of times but if `/dev/urandom` already gave us 4 consecutive bytes of `0` I don't know that it's going to give us anything useful.

flihp:
Apologies for all of the churn on this PR. I ran into an issue with tests passing on my dev system while failing in travis-ci. I sorted out a workaround and integrated it into this branch. This change was limited to the TCTI library and should have no material impact on any of the parts reviewed here.

src/util.c
williamcroberts: I'd at least try a fixed number of times and fail in initialization. I'm assuming those asserts turn into abort() calls on debug builds.

PR #573, opened 2019-01-03
Fix building with clang after recent commits

joshuagl:
When building with clang (as we do in the tpm2-tools Travis CI) we observe the
following error:

src/resource-manager.c:988:41: error: missing field 'data' initializer [-Werror,-Wmissing-field-initializers]
    TPMS_CAPABILITY_DATA cap_data = { 0 };


flihp:
LGTM. Might be good to note that your approach guarantees the remainder of the structure is initialized per C99 rules so it's functionally equivalent to the code being replaced and is better aligned with the C spec.

PR #574, opened 2019-01-07
build: update to ax_code_coverage.m4 version 2019.01.06

diabonas:
When using this project in combination with the freshly released Autoconf Archive 2019.01.06, the build fails with a "missing separator" error in the makefile. This is because upstream made a [breaking change](http://git.savannah.nongnu.org/cgit/autoconf-archive.git/commit/m4/ax_code_coverage.m4?id=b5a2d69f7a7e0133733673586231b88464a98d58) by removing `@CODE_COVERAGE_RULES@` and replacing it with `include $(top_srcdir)/aminclude_static.am`, see the [documentation](https://www.gnu.org/software/autoconf-archive/ax_code_coverage.html) for reference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ AUTHORS
 Makefile
 Makefile.in
 aclocal.m4
+aminclude_static.am
 autom4te.cache/
 compile
 config.guess
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,9 +48,9 @@ matrix:
 install:
   - pip install --user cpp-coveralls
   - mkdir ${DEPS} && pushd ${DEPS}
-  - wget https://download.01.org/tpm2/autoconf-archive-2017.09.28.tar.xz
-  - sha256sum autoconf-archive-2017.09.28.tar.xz | grep -q 5c9fb5845b38b28982a3ef12836f76b35f46799ef4a2e46b48e2bd3c6182fa01
-  - tar xJf autoconf-archive-2017.09.28.tar.xz && pushd autoconf-archive-2017.09.28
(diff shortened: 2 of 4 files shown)


diabonas:
Note that the CI will fail until the corresponding change is merged [upstream into tpm2-tss](https://github.com/tpm2-software/tpm2-tss/pull/1238).

flihp:
This is great. Delaying till https://github.com/tpm2-software/tpm2-tss/pull/1238 is merged though.

diabonas:
The upstream PR has been merged, so this should be good to go now. Apparently there is a bug in the new version of `ax_code_coverage.m4`:  the macro[ `AM_DISTCHECK_CONFIGURE_FLAGS` is recursively defined](https://github.com/autoconf-archive/autoconf-archive/blob/90722129f75e09827c4cc67eacbaf4707225d18e/m4/ax_code_coverage.m4#L178), which causes `make distcheck` to fail when `--enable-code-coverage` is used. I worked around this issue by overwriting the faulty definition and will raise the bug upstream.

diabonas:
Update: the bug in `ax_code_coverage.m4` [was fixed upstream](https://github.com/autoconf-archive/autoconf-archive/pull/182), but until the next release of the Autoconf Archive, we will still have to work around it.

Furthermore there was an [issue with older versions of the Autoconf Archive](https://github.com/tpm2-software/tpm2-tss/pull/1238) with https://github.com/tpm2-software/tpm2-tss/pull/1238, which was fixed in https://github.com/tpm2-software/tpm2-tss/pull/1256. This PR contains this fix as well, so it should work for all versions of the Autoconf Archive.

flihp:
Thanks for turning around the fixup so quick @diabonas. Bonus points for pushing a fix upstream to autoconf-archive for the code-coverage fix! Merging now.

PR #582, opened 2019-02-13
ipc-frontend-dbus: only own bus name after intializing dbus_daemon_proxy

jluebbe:
When started via dbus activation, it can happen that a request is
recieved before dbus_daemon_proxy is set by on_get_dbus_daemon_proxy. In
that case, get_pid_from_dbus_invocation has no proxy and fails.

That results in an error on the client side:
```
  ** (process:521): WARNING **: 00:02:34.701: Failed to create connection
  with service: GDBus.Error:com.intel.tss2.Tabrmd.Error.Internal: Failed
  to get client PID
  ERROR: tcti init allocation routine failed for library: "tabrmd"
  options: "(null)"
  ERROR: Could not load tcti, got: "tabrmd"
```

Fix this by registering the name only when the proxy is initialized
already.

I've tested this with a loop that starts tpm2-abrmd and uses tpm2_getrandom to start it via bus-activation.

### DIFF
--- a/src/ipc-frontend-dbus.c
+++ b/src/ipc-frontend-dbus.c
@@ -649,9 +649,17 @@ on_get_dbus_daemon_proxy (GObject      *source_object,
                    "(org.freedesktop.DBus): %s", error->message);
         g_error_free (error);
         self->dbus_daemon_proxy = NULL;
-    } else {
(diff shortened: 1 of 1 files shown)


flihp:
LGTM. The race is obvious with the data you've included. Thanks for writing up such a clear commit message.

PR #583, opened 2019-02-25
ipc-frontend-dbus: only own bus name after intializing dbus_daemon_proxy

flihp:
When started via dbus activation, it can happen that a request is
recieved before dbus_daemon_proxy is set by on_get_dbus_daemon_proxy. In
that case, get_pid_from_dbus_invocation has no proxy and fails.

That results in an error on the client side:
  ** (process:521): WARNING **: 00:02:34.701: Failed to create connection
  with service: GDBus.Error:com.intel.tss2.Tabrmd.Error.Internal: Failed
  to get client PID
  ERROR: tcti init allocation routine failed for library: "tabrmd"
  options: "(null)"
  ERROR: Could not load tcti, got: "tabrmd"

Fix this by registering the name only when the proxy is initialized
already.

Signed-off-by: Jan Luebbe <jlu@pengutronix.de>

### DIFF
--- a/src/ipc-frontend-dbus.c
+++ b/src/ipc-frontend-dbus.c
@@ -635,9 +635,17 @@ on_get_dbus_daemon_proxy (GObject      *source_object,
                    "(org.freedesktop.DBus): %s", error->message);
         g_error_free (error);
         self->dbus_daemon_proxy = NULL;
-    } else {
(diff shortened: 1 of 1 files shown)


flihp:
This is a cherry-pick of #582 over to the master branch. This PR will be followed by bugfix releases from both 1.x and 2.x release series.

Issue #472, opened 2018-06-05
misleading comment with get_test_opts_from_env


flihp:
https://github.com/tpm2-software/tpm2-abrmd/blob/master/test/integration/test-options.c#L80

the commend indicates that the argv is parsed but really we're pulling data from the environment

PR #590, opened 2019-02-27
test: Fix misleading / inaccurate comment in test-options.

flihp:
We don't pull data from 'argv' in this case, we use the environment
instead.


flihp:
this resolves #472

PR #599, opened 2019-03-08
Changelog 2.1.1

flihp:


flihp:
github isn't very good at dealing with merge commits in PRs. Will be pushing this one directly since PR won't allow me to merge w/o a merge commit (and there's already one in there).

Issue #601, opened 2019-03-10
2.1.1: build fails


kloczek:
<pre>[tkloczko@domek tpm2-abrmd-2.1.1]$ make
make  all-am
make[1]: Entering directory &apos;/home/tkloczko/rpmbuild/BUILD/tpm2-abrmd-2.1.1&apos;
/bin/sh ./libtool  --tag=CC   --mode=link gcc -Wall -Wextra -Werror -std=gnu99 -Wformat -Wformat-security -Wno-missing-braces -Wno-cast-function-type -fdata-sections -ffunction-sections -fstack-protector-all -fpic -fPIC -Wstrict-overflow=5 -D_GNU_SOURCE -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -I./src -I./src/include -I./src -I/usr/include/dbus-1.0 -I/usr/lib64/dbus-1.0/include  -I/usr/include/gio-unix-2.0 -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include  -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include  -pthread     -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -flto -Os -Wl,--gc-sections -Wl,--no-undefined -Wl,-z,noexecstack -Wl,-z,now -Wl,-z,relro  -Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -flto -fuse-linker-plugin -o src/tpm2-abrmd src/tabrmd.o -ldbus-1  -lgio-2.0 -lgobject-2.0 -lglib-2.0  -lglib-2.0   -ltss2-sys -ltss2-mu  src/libutil.la -ldl 
[...]

PR #606, opened 2019-03-21
tcti-factory: Add check for null after TCTI initialization.

flihp:
The code checks the return value and assumes that the context has been
set to a non-NULL value if the response code indicates success. Most
static analysis tools don't look down into the 'tcti_util_dynamic_init'
function to see that this is safe however. Adding an explicit check
comes at nearly no cost and makes the code clear. Some cleanup was
required in the unit test that relied on the old behavior as well.


flihp:
this resolves #601

flihp:
Groan scan-build. Current unit test causes scan-build to fail on a false positive.

PR #609, opened 2019-04-04
test: remove start-auth-session.int.c from the XFAIL list

jeffpferreira:
The test creates a single POLICY session and then calls Tss2_Sys_GetCapability()
to report on the number of session handles.  The test failure was caused by
using TPM2_ACTIVE_SESSION_FIRST instead of TPM2_LOADED_SESSION_FIRST and
vice-versa.  LOADED_SESSION_FIRST is defined as HMAC_SESSION_FIRST and
ACTIVE_SESSION_FIRST is defined as POLICY_SESSION_FIRST.  During the test we
should expect 0 of the former and 1 of the latter.

Signed-off-by: Jeffrey Ferreira <jeffpferreira@gmail.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -71,8 +71,7 @@ TESTS_INTEGRATION_NOHW = test/integration/tcti-connect-multiple.int
 # empty init for these since they're manipulated by conditionals
 TESTS =
 noinst_LTLIBRARIES =
-XFAIL_TESTS = \
-    test/integration/start-auth-session.int
+XFAIL_TESTS =
 TEST_EXTENSIONS = .int
 AM_TESTS_ENVIRONMENT = \
     TEST_FUNC_LIB=$(srcdir)/scripts/int-test-funcs.sh \
--- a/test/integration/start-auth-session.int.c
+++ b/test/integration/start-auth-session.int.c
@@ -148,9 +148,9 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
     }
     g_info ("StartAuthSession for TPM_SE_POLICY success! Session handle: "
             "0x%08" PRIx32, session_handle);
-    handles_count (sapi_context, TPM2_LOADED_SESSION_FIRST, &count);
-    g_assert (count == 1);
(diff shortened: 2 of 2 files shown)


jeffpferreira:
After further investigation my code changes do allow the test to pass, but it looks like my description of the fix is not quite right. The interactions between the tpm2-abrmd daemon and the commands in the test are more nuanced than I originally thought. 

After the tpm2-abrmd processes a command, say `TPM2_StartAuthSession` for example, it appears to do some [post-processing of the handles](https://github.com/tpm2-software/tpm2-abrmd/blob/a8b6cdf0f671b29dda39aea90d085b08f0700fbc/src/resource-manager.c#L785) created during the command. The post processing includes `Tss2_Sys_ContextSave` of any session referenced in the command. This has the effect of moving the session just created from LOADED to ACTIVE. The same seems to be true for post processing of `TPM2_ContextLoad` of a session except that tpm2-abrmd keeps its own state about which sessions are loaded vs saved.

This post-processing behavior means that during the entire start-auth-session.int.c test, LOADED_SESSION == 0 and ACTIVE_SESSION == 1, except for the final check after the loaded session has been flushed where both counts are 0.

All that being said, maybe someone could comment on whether start-auth-session is a valid test as written. Perhaps it makes sense to move this test to [tpm2-tss](https://github.com/tpm2-software/tpm2-tss) where we would expect LOADED and ACTIVE counts to change with each call to `TPM2_ContextSave` and `TPM2_ContextLoad`.

flihp:
Take a look at section 3.15 from the relevant spec: https://trustedcomputinggroup.org/wp-content/uploads/TSS_2p0_TAB_ResourceManager_v1p0_r18_04082019_pub.pdf. But generally you're headed in the right direction. If memory serves this test was intended to cover the "virtualization" of session handles. At the time though, reconstructing the appropriate responses to the `GetCapability` command was incredibly arduous since `libtss2-mu` wasn't yet available. So I left the test as `XFAIL` intending to get back to it "later" ... and it's definitely "later".

Since the patch attached to this PR isn't valid I'm closing it. But this is definitely a gap. Created #612 to track.

jeffpferreira:
Thanks for clarifying

Issue #618, opened 2019-05-03
make main module easier to test


flihp:
The code in tabrmd.c is fairly well tested based on the numbers from codecov but nearly all error cases in this module are untested. Testing these cases requires unit tests but having all of this code in a single module makes that impossible since the `main` function is in there as well.

This issue is tracking the disaggregation of this module into however many new modules are required.

flihp:
see #614

PR #619, opened 2019-05-06
Tabrmd options refactor

flihp:
progress on #618

### DIFF
--- a/src/tabrmd-options.c
+++ b/src/tabrmd-options.c
@@ -104,7 +104,7 @@ tcti_conf_parse (gchar *combined_conf,
  * Then we do a bit of sanity checking and setting up default values if
  * none were supplied.
  */
-void
+gboolean
 parse_opts (gint argc,
             gchar *argv[],
             tabrmd_options_t *options)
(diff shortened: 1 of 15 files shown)


src/tabrmd-options.c
```diff
@@ -71,11 +71,11 @@ tcti_conf_parse (gchar *combined_conf,
     }
     if (combined_conf == NULL) {
         g_debug ("%s: combined conf is null", __func__);
-        return TRUE;
+        return FALSE;
```
DomiNic5787: Why is this set to FALSE now. This causes an error, when starting tpm2-abrmd without setting a tcti.
This behavior does not fit to the documentation:
```
If this option is omitted (or a NULL string provided) then a default TCTI is
used in it's default configuration.
```

src/tabrmd-options.c
```diff
@@ -71,11 +71,11 @@ tcti_conf_parse (gchar *combined_conf,
...
     }
     if (strlen (combined_conf) == 0) {
         g_debug ("%s: combined conf is the empty string", __func__);
-        return TRUE;
+        return FALSE;
```
DomiNic5787: see comment above

src/tabrmd-options.c
DomiNic5787: see #625

PR #626, opened 2019-05-14
tabrmd-options: Fix bug introduced in c5e66b78de.

### LINKED ISSUE #625: tpm2_abrmd failing with "bad tcti conf string"
Starting tpm2_abrmd is failing with `bad tcti conf string`, when no tcti is set.

Behavior does not fit to documentation:
```
If this option is omitted (or a NULL string provided) then a default TCTI is
used in it's default configuration.
```

Possible root cause: #619

flihp:
The comment above this function clearly states the reason for these
return values. This was however missed in this refactoring. This fix
includes a small change to free the GOptionContext as early as possible.
This allows us to eliminate the final conditional returning the status
from the last function call directly.

This allows us to eliminate one of the unit tests without affecting the
coverage metrics.

This resolves #625 

### DIFF
--- a/src/tabrmd-options.c
+++ b/src/tabrmd-options.c
@@ -65,17 +65,17 @@ tcti_conf_parse (gchar *combined_conf,
 
     g_debug ("%s", __func__);
     if (tcti_filename == NULL || tcti_conf == NULL) {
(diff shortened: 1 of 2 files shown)


flihp:
If you have the time @DomiNic5787 I've added you as a reviewer. Figure this makes sens since you identified the issue.

PR #630, opened 2019-05-21
build: remove D-Bus dependency

diabonas:
According to `ldd -u -r` D-Bus is not needed and everything compiles and runs file without it. Additionally remove `TCTI_DEVICE_CFLAGS`/`TCTI_DEVICE_LIBS` and `TCTI_SOCKET_CFLAGS`/`TCTI_SOCKET_LIBS` since they aren't defined anywhere.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,7 +10,7 @@ unit-count: check
 
 AM_CFLAGS = $(EXTRA_CFLAGS) \
     -I$(srcdir)/src -I$(srcdir)/src/include -I$(builddir)/src \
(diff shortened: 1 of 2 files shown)


flihp:
Thanks @diabonas, this is great cleanup. Since we're using the glib / gio dbus interfaces all of the dbus dependencies are satisfied by glib. Agree: no need to include dbus explicitly.

The explicit dependencies on the TCTIs is a hold over from the bad old days before the dynamic loading protocol was available. Also good cleanup.

PR #631, opened 2019-05-22
pkg-config: sort out dependencies

diabonas:
As in https://github.com/tpm2-software/tpm2-tss/pull/1417, move the dependencies from `Requires` to `Requires.private` and add further necessary dependencies according to the Makefile.

Like https://github.com/tpm2-software/tpm2-tss/pull/1417, the first commit breaks the public API. However the impact is much smaller than for tpm2-tss since libtss2-tabrmd-tcti is usually not linked, but loaded with `dlopen()`. I found no projects using the pkg-config file within the tpm2-software organisation, so nothing will break here.

The second commit is completely backwards-compatible since it only adds new necessary dependencies. It is pending #630 to actually remove the uncessary libdbus-1 dependency.

### DIFF
--- a/dist/tss2-tcti-tabrmd.pc.in
+++ b/dist/tss2-tcti-tabrmd.pc.in
@@ -7,6 +7,6 @@ Name: tss2-tcti-tabrmd
 Description: TCTI library for communicating with the TPM2 access broker / resource manager daemon (tabrmd).
 URL: https://github.com/tpm2-software/tpm2-abrmd
 Version: @VERSION@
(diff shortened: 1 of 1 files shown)


flihp:
Wow, my discipline w/r/t linking hasn't been great here. In reviewing this I'm seeing a lot of overlinking in other, unrelated areas. I'm tempted to try to sort these out before merging this PR but since this has already sat in the queue for ~2 weeks and they're not issues introduced by this PR I'm merging this as-is and will create a ticket to track the overlinking issue.

PR #632, opened 2019-05-30
Minor build improvements

alonbl:
Hi,

1. Use AC_USE_SYSTEM_EXTENSIONS and let autoconf detect system pre-processor settings instead of hard-coding them

2. Add --disable-defaultflags to allow downstream to control flags, sync with other tpm2 build systems. This requires for downstream maintainers, I am of Gentoo.

Thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -7,6 +7,7 @@ AC_INIT([tpm2-abrmd],
 AC_CONFIG_MACRO_DIR([m4])
 AC_PROG_CC
 AC_PROG_LN_S
+AC_USE_SYSTEM_EXTENSIONS
 LT_INIT()
 PKG_INSTALLDIR()
 AM_INIT_AUTOMAKE([foreign subdir-objects])
@@ -135,6 +136,7 @@ AM_CONDITIONAL([ENABLE_INTEGRATION],[test "x$enable_integration" = "xyes"])
 
 # preprocessor / compiler / linker flags
 #   these macros are defined in m4/flags.m4
+AX_ADD_FORTIFY_SOURCE
 AX_ADD_COMPILER_FLAG([-Wall])
 AX_ADD_COMPILER_FLAG([-Wextra])
 AX_ADD_COMPILER_FLAG([-Werror])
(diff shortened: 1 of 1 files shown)


alonbl:
@diabonas this is similar issue.

configure.ac
```diff
@@ -151,9 +152,6 @@ AX_ADD_TOOLCHAIN_FLAG([-fstack-protector-all])
 AX_ADD_COMPILER_FLAG([-fpic])
 AX_ADD_COMPILER_FLAG([-fPIC])
 AX_ADD_COMPILER_FLAG([-Wstrict-overflow=5])
-AX_ADD_PREPROC_FLAG([-D_GNU_SOURCE])
-AX_ADD_PREPROC_FLAG([-U_FORTIFY_SOURCE])
-AX_ADD_PREPROC_FLAG([-D_FORTIFY_SOURCE=2])
```
diabonas: AFAIK [`AC_USE_SYSTEM_EXTENSIONS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.64/html_node/Posix-Variants.html) only adds `-D_GNU_SOURCE`. For fortification you need [`AX_ADD_FORTIFY_SOURCE`](https://www.gnu.org/software/autoconf-archive/ax_add_fortify_source.html) from the Autoconf Archive.

alonbl: It seems like it builds nicely this way, are you sure it is needed?

diabonas: It should build fine without fortification, but adding it performs some additional buffer overflow checks, see the [feature_test_macros man page](http://man7.org/linux/man-pages/man7/feature_test_macros.7.html), which seems desirable. According to the [Autoconf Archive documentation](https://www.gnu.org/software/autoconf-archive/ax_add_fortify_source.html) Gentoo has it enabled globally, which might be why you don't see a difference on your system.

Given that this is a hardening preference instead of a hard requirement, it might make sense to put `AX_ADD_FORTIFY_SOURCE` in the `--enable-defaultflags` section.

alonbl: Done

alonbl:
@diabonas it should be ok now

Issue #633, opened 2019-06-05
fix overlinking


flihp:
@diabonas points out some of the bad hygiene that's built up in the build specifically w/r/t linking in #631. Some of this is just bit rot as libraries have been dropped / replaced without dependencies being cleaned up. Some is the result of my ignorance w/r/t autotools stuff in the early days of this project that I never went back and fixed.

This issue is intended to track this cleanup. It will remain open till a manual review of all dependencies (not many) for all project components exposed to users (the daemon & TCTI library).

PR #635, opened 2019-06-06
build: cleanup response-sink unit build rules

flihp:
All of but one of the files listed in the SOURCES variable are already
built as part of libutil. No use in building these again.


flihp:
related to #633

PR #639, opened 2019-06-12
util: replace xor address obfuscation / objid with truncated sha256

flihp:
xor worked fine for this purpose but the crypto folks get itchy if we
use anything other than a standard hash algorithm. Luckily glib has a
sha256 implementation that I didn't know about till now. This function
must truncate the hash to a void* so we increase the probability of a
collision. Since this value is never used for anything other than
debug output it's sufficient for the purpose.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -376,35 +376,6 @@ out:
     return rc;
 }
 
-#define ENTROPY_SRC "/dev/urandom"
-static void* ptr_mask = NULL;
-/*
- * This initialization function is used to initialize state internal to the
- * util module. Use of any function from this module that requires
- * initialization before 'util_init' is called should cause the program to
- * halt via an 'assert'. Currently this init function is only used to get a
- * random value from ENTROPY_SRC for use in the 'objid' function.
- */
-void
-util_init (void)
-{
-    if (ptr_mask != NULL) {
-        g_debug ("%s: already initialized", __func__);
-        return;
-    }
-
-    Random *rand = NULL;
-    int ret;
-    size_t bytes;
-
-    rand = random_new ();
-    ret = random_seed_from_file (rand, ENTROPY_SRC);
-    g_assert (ret != -1);
-    bytes = random_get_bytes (rand, (uint8_t*)&ptr_mask, sizeof (ptr_mask));
-    g_assert (bytes == sizeof (ptr_mask));
-
-    g_clear_object (&rand);
-}
 /*
  * This function is used to mask pointer values before they are displayed
  * in debug output. This allows us to use the address of objects to track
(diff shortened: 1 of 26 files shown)


src/util.c
```diff
@@ -414,10 +385,18 @@ util_init (void)
...
+    if (checksum == NULL) {
+        g_warning ("g_checksum_new failed, returning NULL");
         return NULL;
     }
-    return (void*)((uintptr_t)ptr_mask ^ (uintptr_t)ptr);
+    g_checksum_update (checksum, (guchar*)&ptr, sizeof (ptr));
+    g_checksum_get_digest (checksum, (guint8*)buf, &length);
+    g_checksum_free (checksum);
+
+    return (void*)buf[0];
```
williamcroberts:
```diff
diff --git a/src/util.c b/src/util.c
index 87201c9c5411..0b7efff25eb3 100644
--- a/src/util.c
+++ b/src/util.c
@@ -386,8 +386,11 @@ void*
 objid (const void *ptr)
 {
     GChecksum *checksum;
-    uintptr_t buf [64 / sizeof (uintptr_t)] = { 0, };
-    gsize length = sizeof (buf);
+    union {
+        void *ptr;
+        uintptr_t data;
+    } buf = { 0 };
+    gsize length = sizeof (buf.data);
 
     checksum = g_checksum_new (G_CHECKSUM_SHA256);
     if (checksum == NULL) {
@@ -395,8 +398,8 @@ objid (const void *ptr)
         return NULL;
     }
     g_checksum_update (checksum, (guchar*)&ptr, sizeof (ptr));
-    g_checksum_get_digest (checksum, (guint8*)buf, &length);
+    g_checksum_get_digest (checksum, buf.ptr, &length);
     g_checksum_free (checksum);
 
-    return (void*)buf[0];
+    return buf.ptr;
 }
```

src/util.c
```diff
@@ -414,10 +385,18 @@ util_init (void)
 void*
 objid (const void *ptr)
 {
-    g_assert (ptr_mask != NULL);
+    GChecksum *checksum;
+    uintptr_t buf [64 / sizeof (uintptr_t)] = { 0, };
```
williamcroberts: if sizeof(uintptr_t) is ever > 64, this fails.... uintptr_t buf[0] is valid in GCC, but is discouraged:
```
Declaring zero-length arrays in other contexts, including as interior members of structure objects or as non-member objects, is discouraged. Accessing elements of zero-length arrays declared in such contexts is undefined and may be diagnosed.
```
You would want something like:
```C
(sizeof(uintptr_t) * 8) / sizeof(uintptr_t))
```

src/util.c
williamcroberts: If it complains on the g_checksum_get_digest() you can just union in a guchar array[0]; and use buf.array.

flihp:
All good points. This code just isn't worth it though. It's hardly valuable in debugging complex things (just use gdb) and it's incredibly awkward. Better to just remove this liability.

PR #646, opened 2019-07-02
util: remove use of objid

flihp:
This was originally used to track / trace objects moving through the
daemon. These are largely useless now and tracing / debugging should be
done with gdb and the like.


flihp:
this is an alternative approach to the one taken in #639

flihp:
This removes enough code to move the coverage down ~0.25%. This threshold should be higher.

PR #651, opened 2019-08-22
Add include sys/endian.h and define missing pollflag for FreeBSD

tstruk:


src/tpm2-header.c
```diff
@@ -4,7 +4,6 @@
  * All rights reserved.
  */
 #include <assert.h>
-#include <endian.h>
```
flihp: Why remove `endian.h` here?  This file uses `be16toh` so we need `endian.h` no? It may be pulled in by another include but I'd prefer to include it explicitly.

src/util.h
```diff
@@ -9,9 +9,18 @@
...
 #include <gio/gio.h>
 #include <tss2/tss2_tpm2_types.h>
+#if defined(__FreeBSD__)
+#include <sys/endian.h>
+#endif
 
 #include "control-message.h"
 
+#if defined(__FreeBSD__)
+#ifndef POLLRDHUP
```
flihp: This is only used in the TCTI implementation. Only place that includes `poll.h` at least. This belongs over there.

src/util.h
```diff
@@ -9,9 +9,18 @@
 #include <glib.h>
 #include <gio/gio.h>
 #include <tss2/tss2_tpm2_types.h>
+#if defined(__FreeBSD__)
+#include <sys/endian.h>
```
flihp: Nothing in this module uses `endian.h`.

src/tpm2-header.c
tstruk: There are three files that use be<N>toh: functions src/tpm2-command.c, src/tpm2-header.c, and src/tpm2-response.c. Only tpm2-header.c included it explicitly. The others pulled it via <tss2/tss2_mu.h>. I just made it consistent.

src/util.h
tstruk: Right, I will move it there.

src/util.h
tstruk: As I said, three *.c files use endian functions and they all include src/util.h. I would need to make the same #if defined(__FreeBSD__) in all three *.c files or in one common header file that is included by all *.c files.

src/util.h
flihp: thx

src/tpm2-header.c
flihp: thx for the details

src/util.h
flihp: Apologies for making your repeat yourself. Without a commit message to explain this though, I don't know of any other way to reverse engineer your intent. How about you fill in the associated commit message with data from this thread when you update the `poll` bits of the patch?

tstruk: PR updated

flihp:
@tstruk : do you need a release with this fix?

tstruk:
Yes, but it's not urgent. I included a patch in the port with this change. Once there will be a new, regular  release then I will update the port and drop the patch from it. Thanks for merging it.

PR #662, opened 2019-11-04
Release debug macros

### LINKED ISSUE #640: use AX_IS_RELEASE and AX_CHECK_ENABLE_DEBUG
These are some really useful macros that I didn't know existed till @AndreasFuchsSIT pointed them out. Looks like there are two particular benefits:
1) For debugging I've always added the required flags manually through the environment. Having a configure option would be nice.
2) Dialing up warnings & treating warnings as errors in releases has the potential to backfire if the downstream is using a different / newer tool chain. These should only be enabled for builds from git / not a release tarball.

flihp:
Upstream tpm2-tss has been using these macros for some time. This resolves #640. The only interesting stuff here is that we must now explicitly configure the clang / scan-build build to enable debug per guidance here: https://clang-analyzer.llvm.org/scan-build.html#recommended_debug

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
 matrix:
   include:
     - compiler: clang
-      env: SCANBUILD="scan-build --status-bugs" CFLAGS=-D_REENTRANT CONFIG_EXTRA="--enable-code-coverage --disable-defaultflags" MAKE_TARGET=check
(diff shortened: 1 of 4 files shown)


flihp:
pushed additional fixes to:
- remove use of compiler flags obsoleted by this patch
- prevent conflict between default compiler flags and those enabled by the new debug and release macros

PR #674, opened 2019-12-04
tabrmd-init: Exit with EX_IOERR when IPC fails to connect.

### LINKED ISSUE #670: expose more data w/r/t IPC failures
#667 was caused by failures on the initialization path but other unrecoverable failures are possible and these should provide meaningful exit codes as well. Specifically the IPC mechanism may fail in an unrecoverable way and the exit code from the daemon should reflect this as well. This requires a bit more surgery since these errors don't occur on the initialization path. Instead they'll come from an event generated by an IpcFrontend object. This requires that the signal handler be provided with sufficient data to identify whether or not the failure is recoverable / terminal. The signal handler must then pass this data back to the 'main' after it terminates the GMainLoop.

flihp:
When the IPC mechanism fails to connect the tabrmd-init module receives
an event from the IpcFrontend object and then kills off the GMainLoop.
The current implementation provides no way for the main function to know
why the loop was killed off and so the it exits with a status of 0
(EXIT_SUCCESS) which causes systemd to restart the daemon. This causes a
restart loop.

This patch adds a boolean to the gmain_data_t structure that the
callback sets when the IPC disconnects. This allows the main function to
check for the flag when the GMainLoop terminates and return a meaningful
exit code (EX_IOERR). This exit code will prevent systemd from trying to
restart the daemon.

This resolves #670 

### DIFF
--- a/src/tabrmd-init.c
+++ b/src/tabrmd-init.c
@@ -42,17 +42,27 @@ static gboolean
 signal_handler (gpointer user_data)
 {
     g_info ("handling signal");
-    /* this is the only place the global pointer to the GMainLoop is accessed */
     main_loop_quit ((GMainLoop*)user_data);
 
     return G_SOURCE_CONTINUE;
 }
 
+/*
+ * This function is a callback invoked by the IpcFrontend object
+ * when a 'disconnect' event occurs. It sets a flag in the gmain_data_t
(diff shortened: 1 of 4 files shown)


flihp:
one more time with the test cases fixed!

Issue #673, opened 2019-12-04
revisit systemd restart policy


flihp:
Back when the systemd unit was first written the daemon had a few crashes still and so I set the restart policy to 'always'. These crashes were fixed a long time ago, and users are now reporting some situations #667 #661 etc where having systemd restart the daemon isn't desirable. After implementing a fix to prevent unrecoverable failures in the initialization code #669 it's becoming clear that the majority of the errors (all?) reported aren't recoverable. Getting meaningful exit codes in these situations is necessary but it raises a valid question as to whether the `always` restart policy is appropriate (IMHO it's not).

Suggestions from @martinezjavier in #669 https://github.com/tpm2-software/tpm2-abrmd/pull/669#issuecomment-561111842 should be considered.

flihp:
Could use some input from the perspective of a distro / packager @martinezjavier. I've got the daemon returning meaningful exit codes for the following situations:
1) TCTI init fails (including device TCTI failing on account of a missing device node)
2) TPM startup fails / initialization commands fail
3) daemon fails to claim name on dbus (insufficient permissions etc)

Currently the systemd unit will silently refuse to restart the daemon in these situations. If we remove the `Restart` stuff from the systemd unit, attempts to start the daemon will now fail loudly in these cases. Adding the `ConditionPathExistsGlob` to check for the TPM device node will prevent loud failures when the device TCTI can't be initialized, but all of the other error conditions above will cause systemd to fail loudly. So my qeustion to you is: when do you want `systemctl start tpm2-abrmd` to fail silently when when do you want it to fail for real?

flihp:
Moving this out of 2.3.1 to keep it from blocking the release

martinezjavier:
@flihp sorry I was busy with other stuff so this fell off my radar. My opinion is that we never should try to restart the daemon. My rationale is that in most cases it will fail due the fact that a TPM is not present in the machine.

So adding the `ConditionPathExistsGlob` would be enough in most situations.

For the other failures I think is OK to fail loudly and write into the journal, but systemd shouldn't try to restart it since a) it's likely that will fail again and b) it's D-Bus activated anyways so if there's really a client for the daemon then systemd will try to start it again due a D-Bus client call.

I'll propose a PR with this two changes, feel free to chime in and suggest any changes that you see appropriate.

PR #676, opened 2019-12-14
A couple of fixes for the tpm2-abmrd service unit

martinezjavier:
This pull-request makes some changes to the tpm2-abrmd service unit as discussed in issue #673.

It does two things: check that a TPM character device exists to start the service and don't attempt to restart the service since it may lead to a restart loop.


flihp:
Thx for this. I'll resolve the conflict from #675 manually.

PR #678, opened 2019-12-17
Update SELinux policy for network TCTI connections

trevor-vaughan:
Updated the current SELinux policy with an SELinux boolean
`tpm_abrmd_can_network_connect` which uses the `tcs_port_t` restriction
for external system connectivity.

Fixes #657

Signed-off-by: Trevor Vaughan <tvaughan@onyxpoint.com>

PR #680, opened 2019-12-18
selinux: Allow name resolve and optional unreserved port name_connect.

flihp:
To use the mssim TCTI the daemon must be able to connect to it. The
simulator binds to ports 2321 and 2322 by default. Neither of these are
reserved ports and so they're labeled 'unreserved_port_t' (the default
for ports 1024-32767, and 61001-65535). This commit adds the required
allow rule in a tunable block guarded by a new boolean:
'tabrmd_can_network_connect'. This boolean is disabled by default so use
of mssim TCTI will require enabling it.

Additionally the daemon may need to resolve network names as part of
connecting to the simulator. The available interface from sysnet cannot
be used from within the 'tunable_policy' block (no clue why) and since
this is a largely innocuous thing for a system daemon to do we place it
outside the tunable block.

### DIFF
--- a/selinux/tabrmd.te
+++ b/selinux/tabrmd.te
@@ -1,9 +1,11 @@
-policy_module(tabrmd, 0.0.1)
+policy_module(tabrmd, 0.0.2)
 
 ########################################
 #
 # Declarations
 #
+
+gen_tunable(`tabrmd_connect_all_unreserved', false)
 
 type tabrmd_t;
 type tabrmd_exec_t;
@@ -13,6 +15,7 @@ allow tabrmd_t self:unix_dgram_socket { create_socket_perms };
 
 dev_rw_tpm(tabrmd_t)
 logging_send_syslog_msg(tabrmd_t)
+sysnet_dns_name_resolve(tabrmd_t)
 
 optional_policy(`
     dbus_stub()
@@ -20,3 +23,6 @@ optional_policy(`
     allow system_dbusd_t tabrmd_t:unix_stream_socket rw_stream_socket_perms;
 ')
 
+tunable_policy(`tabrmd_connect_all_unreserved',`
+    corenet_tcp_connect_all_unreserved_ports(tabrmd_t)
+')


flihp:
@lukehinds - any chance I can get a review from you? I took a different approach to the first PR after looked at it more closely (my mistake). The part I'm unsure of is allowing the daemon to resolve names by default. Seems harmless to me but feel free to disagree.

flihp:
#678

lukehinds:
Thanks for the PR @flihp , it looks good to me, but I am not a SME when it comes to SELinux. 

@wrabcak I see you have contributed before and know SELinux well, does this approach look safe to you?

selinux/tabrmd.te
```diff
@@ -13,10 +15,16 @@ allow tabrmd_t self:unix_dgram_socket { create_socket_perms };
...
+sysnet_dns_name_resolve(tabrmd_t)
 
 optional_policy(`
     dbus_stub()
     dbus_system_domain(tabrmd_t, tabrmd_exec_t)
     allow system_dbusd_t tabrmd_t:unix_stream_socket rw_stream_socket_perms;
 ')
 
+tunable_policy(`tabrmd_can_network_connect',`
+    gen_require(`type unreserved_port_t;')
```
wrabcak: This line could be removed

selinux/tabrmd.te
```diff
@@ -13,10 +15,16 @@ allow tabrmd_t self:unix_dgram_socket { create_socket_perms };
...
+
+    allow tabrmd_t unreserved_port_t:tcp_socket name_connect;
```
wrabcak: This line should be replaced by: corenet_tcp_connect_all_unreserved_ports(tabrmd_t)

selinux/tabrmd.te
```diff
@@ -1,10 +1,12 @@
-policy_module(tabrmd, 0.0.1)
+policy_module(tabrmd, 0.0.2)
 
 ########################################
 #
 # Declarations
 #
 
+gen_tunable(`tabrmd_can_network_connect', false)
```
wrabcak: I would rename the name of boolean to: tabrmd_connect_all_unreserved

wrabcak:
I proposed some minor changes, but yes, this could be allowed. 

Thanks,
Lukas.

trevor-vaughan:
Hmm...is allowing unrestricted port connections a good idea for something this critical to system security?

I suppose it's not a huge issue, but this is why I initially co opted the `tcs_port_t` port. https://github.com/tpm2-software/tpm2-abrmd/pull/678#issuecomment-567205111

It could always be tightened down later if a new port is introduced into core.

wrabcak:
> Hmm...is allowing unrestricted port connections a good idea for something this critical to system security?
> 

Yes, you're right, for this purpose we have SELinux booleans. This allow rule is by default turned off, and system administrator needs to allow it before make allow rule active. 

> I suppose it's not a huge issue, but this is why I initially co opted the `tcs_port_t` port. [#678 (comment)](https://github.com/tpm2-software/tpm2-abrmd/pull/678#issuecomment-567205111)
> 
> It could always be tightened down later if a new port is introduced into core.

Related to tcs_port_t. This could be enabled by default, or by another boolean which is less "dangerous". 

Thanks,
Lukas.

flihp:
I've incorporated the feedback (thanks!) from @wrabcak but I'm stuck in an airport and can't access my Fedora VM. Will get this updated, tested etc when I'm back, hopefully tomorrow :crossed_fingers:

wrabcak:
No problem, 
The best would be do some functional tests if there are no more SELinux denials.

flihp:
Agree. The relevant code path is the `Init` function in `libtss2-tcti-mssim` so just starting the daemon is enough to get the library to connect to the simulator on the `unlabeled_port_t` type. Makes sense to run a few of the tools from https://github.com/tpm2-software/tpm2-tools for good measure. Good way to verify send / receive permissions too. Works as expected.

PR #682, opened 2019-12-21
ci: disable building the FAPI for tpm2-tss

diabonas:
Like the ESAPI, the FAPI is currently not used an pulls in additional dependencies (curl and json-c), see https://github.com/tpm2-software/tpm2-tools/pull/1877, so disable it when building tpm2-tss.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -3,7 +3,6 @@ extraction:
     prepare:
       packages:
       - autoconf-archive
-      - libssl-dev
     after_prepare:
     - cd "$LGTM_WORKSPACE"
     - mkdir installdir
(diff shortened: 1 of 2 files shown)


diabonas:
The LGTM failure for this PR is expected as building the base commit fails since it doesn't include the LGTM configuration update yet. The important part is that [building the merge commit works](https://lgtm.com/projects/g/tpm2-software/tpm2-abrmd/logs/rev/pr-dbff96e41e6f8b93e6aaad3d937a8d69bd453492/lang:cpp/stage:Build%20merge_374a02282c50d5bdb35cd5558d451cf91bcdfbf6).

PR #683, opened 2019-12-24
(WIP) Use a specific device dependency instead of udev-settle

grawity:
Pulling in udev-settle affects the entire system – it makes the entire boot process block until all events have been processed. This is usually overkill when we only need one specific device.

Note: This is an incomplete change as I still haven't figured out how to properly add the udev rule into `Makefile.am`, but maybe it would be better to do that in the existing .rules file that's part of tpm2-tss instead?

### DIFF
--- a/dist/tpm2-abrmd.service.in
+++ b/dist/tpm2-abrmd.service.in
@@ -1,9 +1,9 @@
 [Unit]
 Description=TPM2 Access Broker and Resource Management Daemon
-After=systemd-udev-settle.service
-Requires=systemd-udev-settle.service
-# This condition is needed when using the device TCTI. If the
-# TCP mssim is used then the condition should be commented out.
+# These settings are needed when using the device TCTI. If the
+# TCP mssim is used then the settings should be commented out.
+After=dev-tpm0.device
+Requires=dev-tpm0.device
 ConditionPathExistsGlob=/dev/tpm*
 
 [Service]


flihp:
This is great. Coordination with upstream tpm2-tss is a good point. Since we've got an RC running now for a bugfix release I'm going to queue this up for consideration after the RC ends.

tstruk:
https://github.com/tpm2-software/tpm2-tss/pull/1586 is merged

tstruk:
BTW don't you want this to be:
After=dev-tpm*.device
Requires=dev-tpm*.device

grawity:
No, wildcards don't work in dependencies – it needs to list an exact unit name. (And would it be "requires all matching units" or "requires at least one matching unit"?)

So I used tpm0 since the count always starts from 0, and I'm kinda assuming that systems with multiple TPMs are exceedingly rare.

grawity:
After reading #673 I'm wondering if I need to re-think this – due to After= being processed earlier than Condition*=, attempting to start it on a system with no TPM will actually wait for timeout rather than quietly skipping the service (i.e. the Condition*= becomes useless).

williamcroberts:
> After reading #673 I'm wondering if I need to re-think this – due to After= being processed earlier than Condition*=, attempting to start it on a system with no TPM will actually wait for timeout rather than quietly skipping the service (i.e. the Condition*= becomes useless).

I'm confused, wouldn't these three make sure that a TPM exists?
```
After=dev-tpm0.device
Requires=dev-tpm0.device
ConditionPathExistsGlob=/dev/tpm*
```

We could probably drop the `ConditionPathExistsGlob` now? Is that right @martinezjavier ?

diabonas:
IIUC the issue is that `After=dev-tpm0.device` makes the service wait until a certain timeout for a TPM to appear, after which it will go into a "dependency failure" state due to `Requires=dev-tpm0.device` not being fulfilled. With the current combination of `After=systemd-udev-settle.service` and `ConditionPathExistsGlob=/dev/tpm*`, the unit waits instead until all devices have been discovered by udev, then checks whether a TPM exists and fails silently if this is not the case.

Personally, I don't mind a loud failure if `tpm2-abrmd.service` is enabled, but no TPM could be found (if the service is enabled on a system, I would assume it has a TPM, and the user should be warned if that is not the case, e.g. because it failed to initialise). However, there might be distributions that enable `tpm2-abrmd.service` by default on all systems, so these would need to find a way to conditionally enable the service, or live with the log spam of failed service startups.

You are totally correct that `ConditionPathExistsGlob` can be removed with the change proposed in this PR, it serves no purpose any more since it is covered by `Requires=dev-tpm0.device`.

williamcroberts:
Moving to PR: https://github.com/tpm2-software/tpm2-abrmd/pull/755

PR #685, opened 2020-01-13
AccessBroker: Gracefully fail if GetCapability doesn't succeed

martinezjavier:
If calling Tss2_Sys_GetCapability() to get the TPM capabilities fail, the
access_broker_get_tpm_properties_fixed() function calls g_error() which
emits a signal that terminates the tpm2-abrmd process.

This leads to a core dump and a crash being reported in the journal. Since
the service is D-Bus activated, the daemon will be started for each D-Bus
message which makes the issue even more annoying for users.

So instead of killing the process, just log a warning and return an error
message to the access_broker_get_tpm_properties_fixed() function callers.


flihp:
thanks @martinezjavier

PR #686, opened 2020-01-14
build: Add configure option to enable SELinux policy build.

### LINKED ISSUE #495: SELinux policy build @ configure time
We've got instructions for building the SELinux policy module in the INSTALL.md file now which is great. Still we can make things easier: something like an autoconf `with` or `enable` parameter. I can't remember which is the right one for this situation though.

flihp:
This commit adds the `--with-refpol` option to the configure script. If
provided (disabled by default) the SELinux policy module is built and
installed by the default make target. The existing functionality
provided by the hand-rolled `Makefile` in the `$(srcdir)/selinux` directory
remains unchanged to preserve backward compatibility but should removed
as much as possible in the future.

Some weirdness in this commit is required due to weirdness in the
SELinux build infrastructure. Specifically we cannot simply include the
SELinux devel Makefile in Makefile.am. Also we cannot invoke the SELinux
devel Makefile from the Makefile generated by autotools in `$(srcdir)` as
it requires the SELinux policy sources be in the directory where make
was invoked. This forces us to invoke `make -C selinux` from the
Makefile in `$(srcdir)` to kick off the Makefile under `$(srcdir)/selinux`.

This resolves #495


flihp:
@martinezjavier : I've tagged you as a reviewer on this since anything I get wrong here will either break the Fedora build or turn into a bug report that will go to RedHat and then trick down to me. I've tried to preserve all of the existing functionality to keep from breaking the CI you all have setup currently but eventually I'd like to move all functionality possible from $(srcdir)/selinux/Makefile to the autotools stuff.

I've tested this build on Fedora 31 and with the `configure` script passed `--with-refpol` the subsequent call to `make && make install` builds the policy module, compresses it with bzip2, and then installs it to `${prefix}/share/selinux`. The only functionality I haven't duplicated is the policy loading stuff which I would argue doesn't belong in the build (possibly a controversial take?).

Any feedback you can provide would be much appreciated. If / when this gets merged I plan to start removing the duplicate functionality remaining in the current hand-rolled `$(srcdir)/selinux/Makefile`.

martinezjavier:
> eventually I'd like to move all functionality possible from $(srcdir)/selinux/Makefile to the autotools stuff.

Yes, we have a `tpm2-abrmd-selinux` package that's using the hand-rolled Makefile but I'll change it to build with the `--with-refpol` instead once we have a release containing this option, so the Makefile can be removed.

> The only functionality I haven't duplicated is the policy loading stuff which I would argue doesn't belong in the build (possibly a controversial take?).

Agreed that this doesn't belong to the build. I admit that was my mistake when adding the Makefile, since that functionality should had been handled in the package instead. I'll do that when changing it to use the `--with-refpol` option.

The changes looks good to me, I just wonder if `refpol` is the correct name for this option. My understanding is that it means "reference policy"?

Maybe `--with-sepolicy` would be easier for users to understand what the option is about? After all I doubt that people would use a different SELinux policy, or if they do then can patch the policy files or contribute upstream if they find issues with it.

flihp:
Thanks for the feedback. Will do the renaming you suggest and get this merged. Cleanup to the `selinux/Makefile` will follow.

flihp:
My approach to this problem has drastically changed since this PR started. Using the Makefile shipped by the selinux repol dev package in the project autotools build broke VPATH builds. Working around this required reimplementing functionality from this Makefile directly in the autotools build. Some of this is in the autoconf script where we discover all of the required binaries, support and interface files from the selinux dev package. The rest are straight forward make rules added to the automake file.

PR #689, opened 2020-02-13
ci: install libgcrypt-dev in CI worker

flihp:
Upstream tpm2-tss reworked their libgcrypt detection in autoconf adding
the use of the AM_PATH_LIBGCRYPT macro. Previously the dependency on
gcrypt was only apparent if the source tree was configured to use it
(instead of the openssl default). Using this macro however causes the
bootstrap process to be dependent on libgcrypt-dev (where the macro
lives). This requires that we install libgcrypt-dev on CI workers even
though we're not even building libtss2-esys (the library that depends on
the crypto stuff).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
       - gnulib
       - lcov
       - libcmocka-dev
+      - libgcrypt-dev
       - libglib2.0-dev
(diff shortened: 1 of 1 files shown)


flihp:
this is causing build failures for PRs

PR #710, opened 2020-03-22
Enable s390x

### LINKED ISSUE #694: enable s390x arch support in ci
Travis is enabling this arch but currently it fails to build the tpm simulator. This looks to be an issue with the build determining the appropriate radix bits for the architecture:

```
LibSupport.h:78:9: error: #error Unable to determine RADIX_BITS from compiler environment
 #       error Unable to determine RADIX_BITS from compiler environment
         ^~~~~
```

I'm pretty sure this is related to some of my work here: https://github.com/microsoft/ms-tpm-20-ref/commit/2140b7523faf89ccb0503fb40ca339855347e2f3

flihp:
This resolves #694

### DIFF
--- a/.ci/enable-s390x.diff
+++ b/.ci/enable-s390x.diff
@@ -1,0 +1,13 @@
+Index: ibmtpm/src/LibSupport.h
+===================================================================
+--- ibmtpm.orig/src/LibSupport.h
++++ ibmtpm/src/LibSupport.h
+@@ -68,7 +68,7 @@
+ #   if defined(__x86_64__) || defined(__x86_64)				\
+     || defined(__amd64__) || defined(__amd64) || defined(_WIN64) || defined(_M_X64) \
+     || defined(_M_ARM64) || defined(__aarch64__) \
+-    || defined(__powerpc64__) || defined(__ppc64__)
++    || defined(__powerpc64__) || defined(__ppc64__) || defined(__s390x__)
+ #       define RADIX_BITS                      64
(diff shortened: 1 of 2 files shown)


flihp:
the s390x build workers on travis-ci are timing out

.ci/enable-s390x.diff
```diff
@@ -0,0 +1,13 @@
+Index: ibmtpm/src/LibSupport.h
+===================================================================
+--- ibmtpm.orig/src/LibSupport.h
++++ ibmtpm/src/LibSupport.h
+@@ -68,7 +68,7 @@
+ #   if defined(__x86_64__) || defined(__x86_64)				\
+     || defined(__amd64__) || defined(__amd64) || defined(_WIN64) || defined(_M_X64) \
+     || defined(_M_ARM64) || defined(__aarch64__) \
+-    || defined(__powerpc64__) || defined(__ppc64__)
++    || defined(__powerpc64__) || defined(__ppc64__) || defined(__s390x__)
```
williamcroberts: isn't this stuff fun....

flihp:
This will require some research to "fix". Would be nice to get a build / test run on real hardware but s390x hardware isn't something you can buy "off the shelf". Best we can probably do is open a ticket with travis-ci / search for others with similar issues.

May be worth moving this to a topic branch hosted here so I don't lose the work.

.ci/enable-s390x.diff
flihp: meh

flihp:
These are hanging / timing out consistently.

PR #725, opened 2020-07-09
build: remove gcrypt

tstruk:


tstruk:
This depends on https://github.com/tpm2-software/tpm2-tss/pull/1781

PR #733, opened 2020-09-24
tabrmd-init: fix memory leaks for tpm2 and sapi_context

whzhe51:
Signed-off-by: whzhe <wanghongzhe@huawei.com>


williamcroberts:
Dropping in favor of: https://github.com/tpm2-software/tpm2-abrmd/pull/752

PR #734, opened 2020-10-27
test: add support for swtpm simulator and switch CI to it

diabonas:
`libtss2-tcti-swtpm` is available in tpm2-tss 3.0.0. Keep support for `libtss2-tcti-mssim` to allow running the integration tests with older tpm2-tss versions as well, but prefer swtpm if it is available.


williamcroberts:
Sweet, I can rebase this ontop of the switch to Github Actions, and then drop the IBM simulator from our docker images.

PR #735, opened 2020-10-27
ci: fix LGTM configuration

diabonas:
The analysis currently [fails](https://lgtm.com/projects/g/tpm2-software/tpm2-abrmd/logs/rev/01a3f6297261085b8ff5b6deb3fbacf14b5e50d6/lang:cpp/stage:Build%20child_01a3f6297261085b8ff5b6deb3fbacf14b5e50d6) due to the missing dependency `libssl-dev`.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -3,6 +3,7 @@ extraction:
     prepare:
       packages:
       - autoconf-archive
+      - libssl-dev
     after_prepare:
     - cd "$LGTM_WORKSPACE"
     - mkdir installdir


diabonas:
The [LGTM build failure](https://lgtm.com/projects/g/tpm2-software/tpm2-abrmd/logs/rev/pr-4c1c2b5af574be109b414463e6f9345a7d33caea/lang:cpp/stage:Build%20master_01a3f6297261085b8ff5b6deb3fbacf14b5e50d6) is expected since the current master cannot be built successfully, so the comparison between the old and new revision fails. The important part in the linked build log is that building the merge commit succeeds, so LGTM will start working again after merging this PR.

williamcroberts (commented):
Actually the configure --disable-esapi changed to --disable-esys, so when that's disabled we shouldn't need libssl-dev.

williamcroberts:
Should be fixed in:
  - https://github.com/tpm2-software/tpm2-abrmd/commit/a2cf2a1fb3524ccee781091ee924dab803d4b91d

PR #738, opened 2020-11-23
ResourceManager: Avoid double free in resource-manager.c

snits:
Clean up potential double free found by coverity in
resource_manager_load_session_from_handle. If flush_session has been
called, don't call session_list_remove which is already called in
flush_session.

Signed-off-by: Jerry Snitselaar <jsnitsel@redhat.com>


snits:
Looking at the glib code, this probably isn't really an issue. The 2nd trip through g_list_remove shouldn't find the element in the list so there won't be another free. Closing this PR.

PR #739, opened 2020-11-23
ResourceManager: Avoid double free in resource-manager.c

snits:
Clean up potential double free found by coverity in
resource_manager_load_session_from_handle. If flush_session has been
called, don't call session_list_remove which is already called in
flush_session.

Signed-off-by: Jerry Snitselaar <jsnitsel@redhat.com>


snits:
I think the session_list_remove will be harmless because g_list_remove won't find an element in the list to free, but it probably doesn't hurt to avoid it anyways.

PR #746, opened 2021-01-01
CI: fix mulit-arch builds

williamcroberts:
multi-arch builds seem to be failing because SS isn't working in the
containers. So for now, just run the unit tests on non-x86 container
images.

See bug:
  - https://github.com/tpm2-software/tpm2-software-container/issues/44
  - https://github.com/tpm2-software/tpm2-abrmd/issues/745

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -8,7 +8,7 @@ $DOCKER_BUILD_DIR/.ci/get_deps.sh "$(dirname $DOCKER_BUILD_DIR)"
 pushd $DOCKER_BUILD_DIR
 
 SCANBUILD=""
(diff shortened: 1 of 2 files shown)


williamcroberts:
Nope. It looks the the real reason the where failing is that when I pulled in the swtpm patch, swtpm seccomp fails on the multi-arch stuff with:
```
swtpm socket --tpm2 --server port=$sim_port --ctrl type=tcp,port=$((sim_port + 1)) --flags not-need-init --tpmstate dir=$PWD
swtpm: seccomp_load failed with errno 22: Invalid argument
``

If I disable it via --seccomp action=none, I am good.

PR #747, opened 2021-01-02
tabrmd-init: fix memory leaks for tpm2 and sapi_context

williamcroberts:
Signed-off-by: whzhe <wanghongzhe@huawei.com>
Signed-off-by: w00475903 <wanghongzhe@huawei.com>

### DIFF
--- a/src/tabrmd-init.c
+++ b/src/tabrmd-init.c
@@ -99,6 +99,12 @@ gmain_data_cleanup (gmain_data_t *data)
     if (data->loop != NULL) {
         main_loop_quit (data->loop);
     }
+    if (data->tpm2 != NULL) {
+        if (data->tpm2->sapi_context != NULL) {
+            g_free(data->tpm2->sapi_context);
+        }
+        g_clear_object (&data->tpm2);
+    }
 }
 /*
  * This function initializes and configures all of the long-lived objects


src/tabrmd-init.c
```diff
@@ -99,6 +99,12 @@ gmain_data_cleanup (gmain_data_t *data)
     if (data->loop != NULL) {
         main_loop_quit (data->loop);
     }
+    if (data->tpm2 != NULL) {
+        if (data->tpm2->sapi_context != NULL) {
+            g_free(data->tpm2->sapi_context);
+        }
+        g_clear_object (&data->tpm2);
```
pdxjohnny, line 106: The `!= NULL` check above seems to say that we should free the pointer, not the address of the pointer, right?

```suggestion
        g_clear_object (data->tpm2);
```

src/tabrmd-init.c
```diff
@@ -99,6 +99,12 @@ gmain_data_cleanup (gmain_data_t *data)
     if (data->loop != NULL) {
         main_loop_quit (data->loop);
     }
+    if (data->tpm2 != NULL) {
+        if (data->tpm2->sapi_context != NULL) {
+            g_free(data->tpm2->sapi_context);
```
pdxjohnny, line 104: nit: Should there be a space here?

```suggestion
            g_free (data->tpm2->sapi_context);
```

pdxjohnny, line 104: This doesn't seem to follow the same pattern as

https://github.com/tpm2-software/tpm2-abrmd/blob/01a3f6297261085b8ff5b6deb3fbacf14b5e50d6/src/tpm2.c#L89-L92

williamcroberts:
Dropping in favor of: https://github.com/tpm2-software/tpm2-abrmd/pull/752

Issue #704, opened 2020-03-13
tpm2-abrmd.service depends on deprecated systemd-udev-settle.service


diabonas:
Since #675 `tpm2-abrmd.service` depends on `systemd-udev-settle.service`. Unfortunately this service [has been deprecated since systemd version 243](https://github.com/systemd/systemd/pull/12267) and produces the following warning when pulled in:
```
udevadm[1825]: systemd-udev-settle.service is deprecated.
udevadm[1825]: Hint: please fix tpm2-abrmd.service not to pull it in.
```
Not sure what the alternative is, there [seems to be no direct replacement for this service](https://www.freedesktop.org/software/systemd/man/systemd-udev-settle.service.html). Possibly revert #675 and check the device ownership in the daemon instead?

(Reported by @Foxboron)

williamcroberts:
I wonder if we could add the service and tell it not to start and have the udev rule trigger it? I need to find someone who is familiar with this.

williamcroberts:
> I wonder if we could add the service and tell it not to start and have the udev rule trigger it? I need to find someone who is familiar with this.

but then we'd have a weird dependency between the tpm2-tss udev rules and tpm2-abrmd....

williamcroberts:
I wonder if the proper way is to use BindsTo in the service. I wonder if that trigger would occur after chmod.
```
[Unit]
<...>
BindsTo=<DEVICE UNIT HERE>.device
<...>
After=<DEVICE UNIT HERE>.device
```

williamcroberts:
Hrmm, or maybe a `ExecStartPre` where we perform the correct file permission modifications. This way it's cleanly decoupled from udev but not interfering with it.

PR #754, opened 2021-01-14
tpm2-abrmd.service: remove udev settle

williamcroberts:
udev settle has been deprecated since systemd version 243[1]. Thus we
need to remove its use and their is no direct replacement. The
requirement on waiting for udev, was so that FS permission rules are run
and the device nodes are accessible to the tpm2-abrmd service. Rather
than require udev, the default configuration of tpm2-abrmd would,
eventually, be to open device /dev/tmp0. Folks needing to configure
another device/tcti would need to alter the service options and the
pre exec options using systemd's builtin override system.

1. https://github.com/systemd/systemd/pull/12267

Fixes: #704

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,6 +19,11 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 AX_RECURSIVE_EVAL([$sbindir], [SBINDIR])
 AC_SUBST([SBINDIR])
 
+AC_PATH_PROG([CHMOD], [chmod])
+AC_SUBST([CHMOD])
+AC_PATH_PROG([CHOWN], [chown])
+AC_SUBST([CHOWN])
+
 AC_CONFIG_FILES([Makefile dist/tss2-tcti-tabrmd.pc dist/tpm2-abrmd.service dist/tpm2-abrmd.preset])
 
 # propagate configure arguments to distcheck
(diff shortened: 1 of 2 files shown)


williamcroberts:
@martinezjavier and @diabonas , this was my thoughts on getting away from udev settle. Thoughts?

williamcroberts:
Their is also this approach:
  - https://github.com/tpm2-software/tpm2-abrmd/pull/683
  
I was trying to figure out how the TAG and those device units worked together and that PR does clarify that for me, however it
requires udev rules to be in place, so we would have to bump the minimum TSS version, it would require at least 2.4.0. I'm fine with that and I like this approach over mine.

williamcroberts:
I'm in favor of #683, I didn't understand it until I spent time reading on systemd today. But I'll gather everyone's feedback and hopefully we can move forward and get this all ready for another release.

diabonas:
I'd also prefer #683, mainly because users or distributions might in theory want to use different ownership and permissions for `/dev/tpm0` and this is nicely encapsulated in the udev rules shipped by tpm2-tss, also it's probably be the more "systemd" way of doing things.

#683 is also how we do it in [tpm2-totp](https://github.com/tpm2-software/tpm2-totp/blob/9bcfdcbfdd42e0b2e1d7769852009608f889631c/dist/plymouth-tpm2-totp.service.in#L3-L4) to make sure the service is started only after the TPM becomes available in the boot process, which is tested to work well in systemd-based initramfs images created by mkinitcpio on Arch Linux.

williamcroberts:
Closing in favor of #683

PR #769, opened 2021-03-09
Add missing install deps

williamcroberts:
Related to:
https://github.com/tpm2-software/tpm2-tss/issues/2016

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,6 +12,7 @@ required:
 * pkg-config
 * glib and gio 2.0 libraries and development files
 * libtss2-sys, libtss2-mu and TCTI libraries from https://github.com/tpm2-software/tpm2-tss
+* dbus
 
 **NOTE**: Different GNU/Linux distros package glib-2.0 differently and so
 additional packages may be required. The tabrmd requires the GObject and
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -12,6 +12,9 @@ required:
 * pkg-config
 * glib and gio 2.0 libraries and development files
 * libtss2-sys, libtss2-mu and TCTI libraries from https://github.com/tpm2-software/tpm2-tss
+* systemd:
+  - systemd-sysusers
```
tstruk, line 16: abrmd doesn't use or need systemd-sysusers

tstruk (changes requested):

INSTALL.md
```diff
@@ -12,6 +12,7 @@ required:
 * pkg-config
 * glib and gio 2.0 libraries and development files
 * libtss2-sys, libtss2-mu and TCTI libraries from https://github.com/tpm2-software/tpm2-tss
+* dbus
```
tstruk, line 15: This is a runtime dependency required for integration test.
Should be listed below with the other test suite dependencies.

tstruk (changes requested):

INSTALL.md
williamcroberts: It's more than just that it comes with dbus-run-session, you need dbus for the whole IPC mechanism. Without dbus it won't work right? So you need dbus for the IPC daemon.

```bash
apt show dbus
<snip>
This package contains the D-Bus daemon and related utilities.
<snip>
```

tstruk: yes, you're right

tstruk (approved):

PR #770, opened 2021-03-15
tcti-tabrmd: close extra fds

### LINKED ISSUE #764: tools cannnot connect to abrmd
I suspect I'm missing some install step.  

The dbus-send DBus.ListNames returns       string "com.intel.tss2.Tabrmd"
The Introspect returns   <interface name="com.intel.tss2.TctiTabrmd">

Thise look OK.

I set TPM2TOOLS_TCTI "tabrmd:bus_name=com.intel.tss2.Tabrmd"

But every command returns:

** (process:631): CRITICAL **: 15:09:37.628: CreateConnection expected to return 1 handles, received 2
WARNING:tcti:src/tss2-tcti/tctildr.c:79:tcti_from_init() TCTI init for function 0x7f3210d0be30 failed with a0001 
WARNING:tcti:src/tss2-tcti/tctildr.c:109:tcti_from_info() Could not initialize TCTI named: tcti-abrmd 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: tabrmd 
ERROR:tcti:src/tss2-tcti/tctildr.c:416:Tss2_TctiLdr_Initialize_Ex() Failed to instantiate TCTI 
ERROR: Could not load tcti, got: "tabrmd:bus_name=com.intel.tss2.Tabrmd"

tstruk:
On some platform, eg. MS Windows WSL the g_dbus_proxy_call_with_unix_fd_list_sync()
function behaves differently than on native Linux host and returns two fd handles
instead of one. Need to take care of the extra one and close in manually here so
it won't leak.

Fixes #764

### DIFF
--- a/src/tcti-tabrmd.c
+++ b/src/tcti-tabrmd.c
@@ -558,10 +558,25 @@ tcti_tabrmd_connect (TSS2_TCTI_CONTEXT *context)
     }
     gint num_handles = g_unix_fd_list_get_length (fd_list);
     if (num_handles != 1) {
-        g_critical ("CreateConnection expected to return 1 handles, received %d",
-                    num_handles);
-        rc = TSS2_TCTI_RC_GENERAL_FAILURE;
-        goto out;
+        g_debug ("CreateConnection expected to return 1 handles, received %d",
+                 num_handles);
+        if (num_handles == 2) {
+			/* On some platform, eg. MS Windows WSL the
+			 * g_dbus_proxy_call_with_unix_fd_list_sync() function behaves
+			 * differently than on native Linux host and returns two fd handles
+			 * instead of one. Need to take care of the extra one
+			 * and close in manually here so it won't leak.
+			 */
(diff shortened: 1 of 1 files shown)


tstruk:
This was mainly to enable MS WSL, but it seams WSL has its own problems with dbus, so we don't need this workaround on the main branch.
The reason seams to be that Windows doesn't support AF_UNIX sockets natively, WSL just re-maps them as WINSOCK TCP on the lo interface (127.0.0.1), but this causes trouble. Some projects patch their code to support/enable WSL [1][2]
Maybe a separate branch/fork of abrmd to support WSL will be created in the future.

[1] https://github.com/pulseaudio/pulseaudio/compare/v12.2...Martin1994:v12.2-win32-afunix
[2] https://www.reddit.com/r/bashonubuntuonwindows/comments/bq31sl/launch_gui_applications_in_wsl_with_unix_domain/

PR #776, opened 2021-03-19
Bump tcti-abrmd so version

tstruk:
Bump tcti-abrmd.so version after changing the dbus CreateConnection definition

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -281,8 +281,14 @@ src_libtss2_tcti_tabrmd_la_LIBADD = $(GIO_LIBS) $(GLIB_LIBS) $(TSS2_MU_LIBS) \
 if !ENABLE_ASAN
 UNDEFINED_SYMS := -Wl,--no-undefined
 endif
-src_libtss2_tcti_tabrmd_la_LDFLAGS = -fPIC $(UNDEFINED_SYMS) -Wl,-z,nodelete -Wl,--version-script=$(srcdir)/src/tcti-tabrmd.map
 src_libtss2_tcti_tabrmd_la_SOURCES = src/tcti-tabrmd.c src/tcti-tabrmd-priv.h $(srcdir)/src/tcti-tabrmd.map
+src_libtss2_tcti_tabrmd_la_LDFLAGS = -fPIC $(UNDEFINED_SYMS) -Wl,-z,nodelete
+if HAVE_LD_VERSION_SCRIPT
+src_libtss2_tcti_tabrmd_la_LDFLAGS +=  -Wl,--version-script=$(srcdir)/src/tcti-tabrmd.map
+endif # HAVE_LD_VERSION_SCRIPT
+
+# Bump so version after changing the dbus CreateConnection definition
+src_libtss2_tcti_tabrmd_la_LDFLAGS += -version-info 1:0:0
 
 src_tpm2_abrmd_LDADD   = $(GIO_LIBS) $(GLIB_LIBS) $(PTHREAD_LIBS) \
     $(TSS2_SYS_LIBS) $(TSS2_TCTILDR_LIBS) $(libutil)
--- a/configure.ac
+++ b/configure.ac
@@ -24,6 +24,8 @@ AC_CONFIG_FILES([Makefile dist/tss2-tcti-tabrmd.pc dist/tpm2-abrmd.service dist/
 # propagate configure arguments to distcheck
 AC_SUBST([DISTCHECK_CONFIGURE_FLAGS],[$ac_configure_args])
 
+gl_LD_VERSION_SCRIPT
+
 AX_PTHREAD([], [AC_MSG_ERROR([requires pthread])])
 AC_ARG_ENABLE([unit],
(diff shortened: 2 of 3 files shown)


tstruk:
@diabonas I would like to ask your opinion on this.
What has changed recently is that the interface definition for CreateConnection dbus call has one less param:
https://github.com/tpm2-software/tpm2-abrmd/commit/7d5a73dd879a9fc4280f4d39a80496f1c8d23ad1
This doesn't change the ABI interface between applications and tcti-abrmd.so, but rather the interface between tcti-abrmd.so, which is used by the clients to talk to tpm2-abrmd daemon, and tpm2-abrmd itself.
What would be the best way to make sure that these are in sync?

diabonas:
Hm, so the DBus interface that the `tpm2-abrmd` daemon provides and `libtss2-tcti-tabrmd.so` consumes changes, correct? In that case, I wouldn't bump the soname of `libtss2-tcti-tabrmd.so`, for the following reasons:

- The `tpm2-abrmd` daemon and `libtss2-tcti-tabrmd.so` are supposed to be distributed together (by this project), so the two should be guaranteed to be updated in lockstep anyway by any sane packaging system.
- Bumping the soname is usually used to signal that applications linking against the library need to be rebuilt (and potentially the source code might need to be changed). TCTI libraries are usually `dlopen()`'ed instead, and the soname is often hardcoded to `libtss2-tcti-tabrmd.so.0`, e.g. in [tpm2-tss](https://github.com/tpm2-software/tpm2-tss/blob/29c46f0d27942146a909f7f12c0f8f65c45bec93/src/tss2-tcti/tctildr-dl.c#L38), so all of these would require changing (probably in a backwards-compatible way).
- Even if you are worried about applications bundling their own `libtss2-tcti-tabrmd.so` not in sync with the tpm2-abrmd daemon, bumping the soname wouldn't help for these applications, they would just try to continue using the new DBus interface and fail.

In my opinion no action is required at all regarding this change, since the DBus interface can be considered as an internal interface between the `tpm2-abrmd` daemon and `libtss2-tcti-tabrmd.so`, which are distributed together as the tpm2-abrmd project and shouldn't be separated.

If you are worried about other consumers of the DBus interface (or separate `libtss2-tcti-tabrmd.so` libraries not updated together with the tpm2-abrmd daemon), I'd suggest versioning the [DBus interface name](https://github.com/tpm2-software/tpm2-abrmd/blob/f0a0b8084cdb98886002d1202c6dcf0a3ab88c2e/src/tabrmd.xml#L4) by renaming it e.g. to `com.intel.tss2.TctiTabrmd1`. This way, old consumers would simply not find the new interface any more and bail out with an error as if tpm2-abrmd wasn't installed at all. However, even without this change, applications would error out upon encountering the non-matching interface as well, so I don't think this helps a lot.

tstruk:
> In my opinion no action is required at all regarding this change, since the DBus interface can be considered as an internal interface between the `tpm2-abrmd` daemon and `libtss2-tcti-tabrmd.so`, which are distributed together as the tpm2-abrmd project and shouldn't be separated.

Right, the two should come together so if they go out of sync the CreateConnection() call will complain about wrong param type and fail. No problem there.

> If you are worried about other consumers of the DBus interface (or separate `libtss2-tcti-tabrmd.so` libraries not updated together with the tpm2-abrmd daemon), I'd suggest versioning the [DBus interface name](https://github.com/tpm2-software/tpm2-abrmd/blob/f0a0b8084cdb98886002d1202c6dcf0a3ab88c2e/src/tabrmd.xml#L4) by renaming it e.g. to `com.intel.tss2.TctiTabrmd1`. This way, old consumers would simply not find the new interface any more and bail out with an error as if tpm2-abrmd wasn't installed at all. However, even without this change, applications would error out upon encountering the non-matching interface as well, so I don't think this helps a lot.

I don't think versioning the dbus name is the way to go. The bus name supposed to be a well-known and not change, so I will drop this PR altogether. Thanks for your input.

williamcroberts:
> I don't think versioning the dbus name is the way to go. The bus name supposed to be a well-known and not change, so I will drop this PR altogether. Thanks for your input.

@tstruk speaking of this, phillip left the name in a weird state. Plumbing and some docs say that its configurable as a tcti option parameter when its not. Can you go remove the name= option, Offhand, I would just make it a NOP + WARN so folks setting it for some reason, their stuff still works.

tstruk:
> @tstruk speaking of this, phillip left the name in a weird state. Plumbing and some docs say that its configurable as a tcti option parameter when its not. Can you go remove the name= option, Offhand, I would just make it a NOP + WARN so folks setting it for some reason, their stuff still works.

@williamcroberts It is configurable, it works quite well, and there is a very good use for it.
By using different dbus names we can start multiple instances of tpm2-abrmd, all separated from each other.
It is used extensively by the tools integration tests:
https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/helpers.sh#L343
Your memory is like mine - good, but short ;)

What I meant is that the default "com.intel.tss2.Tabrmd" supposed to be "well-known" and not change.

PR #798, opened 2021-12-22
.github: Add workflow_dispatch to allow manual CI runs & re-runs.

flihp:
Github actions 'timeout' causing builds to be reported as failed. These
can be corrected up by manually re-running the workflow. This requires
the 'workflow_dispatch' event be 'on'.


williamcroberts:
Theirs stuff timing out? Is it "main" The cirrus task or the actual runners on GH actions?

PR #800, opened 2021-12-22
build: Remove use of 'which' to check for existence of gdbus-codegen.

flihp:
This is a redundant check. AC_CHECK_PROG already finds the program on
path and ensures that it is executable.


flihp:
Updated after adding an entry in the CHANGELOG.

williamcroberts (approved):
LGTM

PR #801, opened 2022-01-05
Using composite actions for CI

malikabhi05:
This PR changes the CI slightly by pointing the builds to common custom composite actions. These actions are currently under malikabh/test-ac and must ideally be moved to their own repo in the tpm2-software org.

This PR is currently for review and discussion and is opened as a draft PR. If everything is good then it can be converted into a proper PR and merged.


.github/workflows/main.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         uses:
           malikabhi05/test-ac/runCI@master
         with:
+          PROJECT_NAME: ${{ github.event.repository.name }}
```
williamcroberts, line 20: Nice, and this is a required arg in the composite action (or whatever its called)?

malikabhi05: Yes, this is required. It is used by the ci.sh script for the docker path

PR #819, opened 2022-10-15
resource-manager: bypass command_special_processing when session is p…

whooo:
…rovided

TPM2_GetCapability accepts an audit session, when specified no special processing should be done.
If a session without the audit flag set is passed let the TPM handle the error.

Fixes https://github.com/tpm2-software/tpm2-abrmd/issues/818

### DIFF
--- a/src/resource-manager.c
+++ b/src/resource-manager.c
@@ -1121,8 +1121,10 @@ command_special_processing (ResourceManager *resmgr,
         response = resource_manager_load_context (resmgr, command);
         break;
     case TPM2_CC_GetCapability:
-        g_debug ("processing TPM2_CC_GetCapability");
-        response = get_cap_gen_response (resmgr, command);
+        if (!tpm2_command_has_auths(command)) {
+            g_debug ("processing TPM2_CC_GetCapability");
+            response = get_cap_gen_response (resmgr, command);
+        }
         break;
     default:
         break;
--- a/Makefile.am
+++ b/Makefile.am
@@ -67,7 +67,8 @@ TESTS_INTEGRATION = \
     test/integration/not-enough-handles-for-command.int \
     test/integration/password-authorization.int \
     test/integration/tpm2-command-flush-no-handle.int \
-    test/integration/util-buf-max-upper-bound.int
+    test/integration/util-buf-max-upper-bound.int \
+    test/integration/get-capability-with-session.int
 
 TESTS_INTEGRATION_NOHW = test/integration/tcti-connect-multiple.int
 
@@ -526,6 +527,10 @@ test_integration_tpm2_command_flush_no_handle_int_SOURCES = \
 test_integration_util_buf_max_upper_bound_int_LDADD = $(TEST_INT_LIBS)
 test_integration_util_buf_max_upper_bound_int_SOURCES = \
     test/integration/main.c test/integration/util-buf-max-upper-bound.int.c
+
+test_integration_get_capability_with_session_int_LDADD = $(TEST_INT_LIBS)
+test_integration_get_capability_with_session_int_SOURCES = \
+    test/integration/main.c test/integration/get-capability-with-session.int.c
 
 if WITH_SEPOLICY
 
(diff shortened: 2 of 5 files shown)


src/resource-manager.c
```diff
@@ -1106,8 +1106,13 @@ command_special_processing (ResourceManager *resmgr,
...
 {
     Tpm2Response *response   = NULL;
+    TPM2_CC cc = tpm2_command_get_code (command);
 
-    switch (tpm2_command_get_code (command)) {
+    if (tpm2_command_has_auths(command) && cc == TPM2_CC_GetCapability) {
+      return NULL;
+    }
+
+    switch (cc) {
```
williamcroberts, line 1115: I would prefer we keep the handling of commands within the respective switch statement for the command code, and thus would prefer something like:
```diff
diff --git a/src/resource-manager.c b/src/resource-manager.c
index 6459b1f409e6..45fb36bb99a8 100644
--- a/src/resource-manager.c
+++ b/src/resource-manager.c
@@ -1121,8 +1121,10 @@ command_special_processing (ResourceManager *resmgr,
         response = resource_manager_load_context (resmgr, command);
         break;
     case TPM2_CC_GetCapability:
-        g_debug ("processing TPM2_CC_GetCapability");
-        response = get_cap_gen_response (resmgr, command);
+        if (!tpm2_command_has_auths(command)) {
+            g_debug ("processing TPM2_CC_GetCapability");
+            response = get_cap_gen_response (resmgr, command);
+        }
         break;
     default:
         break;
```

williamcroberts (changes requested):

src/resource-manager.c
```diff
@@ -1106,8 +1106,9 @@ command_special_processing (ResourceManager *resmgr,
                             Tpm2Command     *command)
 {
     Tpm2Response *response   = NULL;
+    TPM2_CC cc = tpm2_command_get_code (command);
 
-    switch (tpm2_command_get_code (command)) {
+    switch (cc) {
```
williamcroberts, line 1111: Don't change this hunk, lets keep the patch as small as possible. and not modify the switch at all.

williamcroberts (changes requested):

williamcroberts (approved):

williamcroberts:
@whooo we should add a test that does that code snippet to ensure we don't break anything?

whooo:
> @whooo we should add a test that does that code snippet to ensure we don't break anything?

Do you have any flow in mind? Only thing I can think of is doing a unit test for `command_special_processing` and passing a Tpm2Command with a session

williamcroberts:
> > @whooo we should add a test that does that code snippet to ensure we don't break anything?
> 
> Do you have any flow in mind? Only thing I can think of is doing a unit test for `command_special_processing` and passing a Tpm2Command with a session

I was thinking an integration test where we start a session and add the handle, but I am OK with a unit test as well.

whooo:
> > > @whooo we should add a test that does that code snippet to ensure we don't break anything?
> > 
> > 
> > Do you have any flow in mind? Only thing I can think of is doing a unit test for `command_special_processing` and passing a Tpm2Command with a session
> 
> I was thinking an integration test where we start a session and add the handle, but I am OK with a unit test as well.

I'll look into creating a integration test which test with an audit session and one which tests with a password session, just need to get familiar with SAPI again

williamcroberts:
> > > > @whooo we should add a test that does that code snippet to ensure we don't break anything?
> > > 
> > > 
> > > Do you have any flow in mind? Only thing I can think of is doing a unit test for `command_special_processing` and passing a Tpm2Command with a session
> > 
> > 
> > I was thinking an integration test where we start a session and add the handle, but I am OK with a unit test as well.
> 
> I'll look into creating a integration test which test with an audit session and one which tests with a password session, just need to get familiar with SAPI again

Their is start_auth_session in common code for testing. It sets up an encrypted session if you need some reference code.

whooo:
Simple integration test added, I only added checks for failure with a password session and the success with an audit session, but don't verify the audit session

williamcroberts (approved):

PR #837, opened 2023-07-17
selinux: call init_nnp_daemon_domain for domain 

jsegitz:
allows for systemd hardenings to be applied to the daemon


jsegitz:
any chance of this being accepted? Is there any change you would like to see?

PR #841, opened 2023-10-18
Add '/dev/tcm0' to default conf

jouyouyun:
A standard similar to TPM has been released in China, called TCM(Trusted Cryptography Module), and its device path is /dev/tcm0. 

The TCM standard is compatible with TPM, and TSS can be used to manage its device path.

### DIFF
--- a/src/tabrmd-init.c
+++ b/src/tabrmd-init.c
@@ -20,6 +20,7 @@
 #include "tabrmd-init.h"
 #include "tabrmd-options.h"
 #include "tabrmd.h"
+#include "tss2_common.h"
 #include "util.h"
 
 /*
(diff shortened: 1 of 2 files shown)


src/tabrmd-init.c
```diff
@@ -47,6 +48,35 @@ signal_handler (gpointer user_data)
     return G_SOURCE_CONTINUE;
 }
 
+/*
+ * This function attempts to initialize the configured device nodes in sequence
+ * and returns once successful, otherwise it continues to try until the end of
+ * the traversal.
+ */
+static gint
+tss2_tctildr_init(const gmain_data_t *data, TSS2_TCTI_CONTEXT **tcti_ctx)
```
williamcroberts, line 57: put this logic in the tpm2-tss backend for the device tcti rather than tabrmd. This way everything using the tcti libraries gets this support. You'll have a dependency on the updated tpm2-tss library and tpm2-abrmd to support this.

williamcroberts (changes requested):

src/tabrmd-init.c
jouyouyun: Thanks for your reply. TCM support has been added to tpm2-tss, see: [tcti: Add '/dev/tcm0'  to default conf]( https://github.com/tpm2-software/tpm2-tss/pull/2696) . When the configuration of tctildr is empty, the default configuration will be automatically used to complete the initialization on the device with only tcm.
However, tpm2-abrmd specifies the configuration when initializing tctildr. If the configuration does not contain tcm, the initialization will fail on the device with only tcm, so this part of the code is added.

williamcroberts: So then why don't we just use NULL for the default and let the tcti-ldr search logic prevail.

jouyouyun: Yes, this seems feasible. I need to verify how to modify it.

jouyouyun: I apologize for the delayed response. After receiving your reply, I reviewed the code and found that in abrmd, the tcti_ctx needs to be constructed using the device path before using functions like tpm2_new/tpm2_init_tpm to complete the initialization.

It is not possible to reuse the TSS initialization function Tss2_Tcti_TctiLdr_Init for TSS2_TCTI_CONTEXT. Therefore, the functionality to parse the TCM device path still needs to be added to abrmd.

williamcroberts: Oh I think I finally understand the problem. If we pass NULL on devices that have TPM, abrmd will attempt tpmrm0 before tpm0, which is not what we want and on tcm devices it would find tcm0, is that right?

If so, make it an array of strings to try rather than parsing on a semi-colon.

williamcroberts (changes requested):

src/tabrmd-init.c
williamcroberts: I also wonder if this is better off in a service file as tpm2-abrmd takes a tcti option. How do you start the service on tcm devices?

jouyouyun: TCM is compatible with the TPM 2.0 specification, the boot method is the same as TPM, only the device path is different.

williamcroberts: > TCM is compatible with the TPM 2.0 specification, the boot method is the same as TPM, only the device path is different.

Yeah exactly, so what you do, you'll see these files under dist, is make tpm2-abrmd start up using the tcti option based on the presence of the TCM device

PR #846, opened 2024-07-15
Add file context entry for /usr/bin/tpm2-abrmd

zpytela:
To comply with the "Unify bin and sbin" Fedora Change [1], file equivalency for the /bin, /sbin, and /usr/sbin paths are now set in selinux-policy to /usr/bin.
This requires follow-up changes in DSP modules, too.

[1] https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin


ca-hu:
any updates here? this is breaking the tpm2-abrmd.service on systems with selinux enabled

see https://bugzilla.suse.com/show_bug.cgi?id=1229047

williamcroberts (approved):

williamcroberts (changes requested):
You need to sign off your changes @zpytela

zpytela:
Signed, sorry that I missed the original request.

Also note this change is temporary until the actual change is made and may be handled in a different way for different supported releases.

williamcroberts (approved):

kkaarreell:
Hello, could you please fix this issue. It is opened for >1 year already.

williamcroberts:
Yes, I can look at this, I need to get the CI working as well.

Issue #855, opened 2026-06-17
Update comparison with the Linux kernel resource manager


Skyb0rg007:
This was prompted by some discussion on the tpm2 [kernel mailing list](https://lore.kernel.org/tpm2/CABppgTcx=Q62MZDm8xKYTZspwAH16vx47B4vVt84Nzjif8ZFvw@mail.gmail.com/T/#u), which was preceded by a discussion in [nixpkgs](https://github.com/NixOS/nixpkgs/pull/509127).

The issue is that the tpm2-abrmd README is very out-of-date, with some of its claims no longer being well-supported.
For example, this repo has not seen any updates to the service code in ~4 years, while the Linux kernel's resource manager has seen updates in that same time period. As a result, some statements such as

> For the short term we're recommending that developers stick to using the tabrmd as the default to get the most stable / widest possible support.

and

> How we get to the ideal future of a single RM in the kernel: our current plan is to prototype various features in user space as a way to get them tested / validated.

do not necessarily seem relevant anymore.

It would be great if the technical differences between `/dev/tpmrm0` and `tpm2-abrmd` could be listed: on the mailing list someone mentioned that the Linux kernel's resource manager doesn't support session un-gapping, and if true would be good to mention in the README.

AndreasFuchsTPM:
That's true. PRs welcome.
Or even better, someone to implement session ungaping for the in-kernel RM.

Skyb0rg007:
Thanks, I'll make a PR that at least removes the misleading statements. Though I'm not a TPM expert, so I can't confidently state what the feature-set of each resource manager is, or make a recommendation on what end-users should use as a default.

PR #856, opened 2026-06-18
Remove unsupported claims from README

Skyb0rg007:
Removes paragraphs from the README that are out-of-date.

Addresses some aspects of #855, but feature comparisons should be written by someone more knowledgeable of tpm2-abrmd and the Linux kernel's resource manager. Such an expert should also include a recommendation on when to use the tpm2-abrmd RM. For now I believe its best to remove the out-of-date parts and have future contributors fill in the missing parts.


moritzbuhl:
Sorry, I did not see this PR. I think non-Linux support is another reason to keep this.
Thank you for cleaning it up, I'll see what to do about other OS support in the section.

Skyb0rg007:
Because the removed sections just mention Linux, is there anything else that's needed here?

moritzbuhl:
Thank you for updating the README!

PR #860, opened 2026-07-30
Update docker images.

moritzbuhl:
Depends on https://github.com/tpm2-software/tpm2-software-container/pull/112

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -38,13 +38,19 @@ fi
 
 export CFLAGS
 
+FALSE_POSITIVES=(
+  'tabrmd-generated.c.*\[deadcode.DeadStores\]'
+  'tabrmd-init.c:185.*\[deadcode.DeadStores\]'
+)
+
 $SCANBUILD ../configure --enable-debug $CONFIG_OPTS $CONFIG_EXTRA
 if [[ "$SCANBUILD" != scan-build* ]]; then
   $SCANBUILD make -j$(nproc) $MAKE_TARGET
 else
   $SCANBUILD make -j$(nproc) $MAKE_TARGET 2>&1 | tee /tmp/build.log || true
+  IGNORE_REGEX=$(IFS='|'; echo "${FALSE_POSITIVES[*]}")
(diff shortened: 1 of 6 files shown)


.ci/docker.run
```diff
@@ -38,13 +38,22 @@ fi
...
+
+# Join array into a single regex alternation: (a|b|c)
+IGNORE_REGEX=$(IFS='|'; echo "${FALSE_POSITIVES[*]}")
+
 $SCANBUILD ../configure --enable-debug $CONFIG_OPTS $CONFIG_EXTRA
 if [[ "$SCANBUILD" != scan-build* ]]; then
   $SCANBUILD make -j$(nproc) $MAKE_TARGET
 else
   $SCANBUILD make -j$(nproc) $MAKE_TARGET 2>&1 | tee /tmp/build.log || true
+  IGNORE_REGEX=$(IFS='|'; echo "${FALSE_POSITIVES[*]}")
```
AndreasFuchsTPM, line 51: Duplicate of l47 ?

src/random.c
```diff
@@ -127,7 +127,6 @@ random_get_bytes (Random    *random,
     uint8_t rand[sizeof (long int)] = { 0, };
 
     g_assert_nonnull (random);
-    assert (random->rand_state);
```
AndreasFuchsTPM, line 130: Why is this needed ?

moritzbuhl: <pre>
  src/random.c: In function 'random_get_bytes':
  src/random.c:130:5: error: the comparison will always evaluate as 'true' for the address of 'rand_state' will never be NULL [-Werror=address]
    130 |     assert (random->rand_state);
        |     ^~~~~~
</pre>
https://github.com/tpm2-software/tpm2-pkcs11/actions/runs/31227147976/job/93023650129?pr=939#step:3:1371

moritzbuhl: I'll add this message to the commit message and push again.

AndreasFuchsTPM (approved):

